PR #15, opened 2020-11-05
fix the restarting of apps and databases

maathor:


test_utilities/src/utilities.rs
```diff
@@ -59,3 +62,44 @@ fn curl_path(path: &str) -> bool {
...
+    let secret_key = aws_secret_access_key();
+    let aws_credentials_envs = vec![
+        ("AWS_ACCESS_KEY_ID", access_key.as_str()),
+        ("AWS_SECRET_ACCESS_KEY", secret_key.as_str()),
+    ];
+
+    let kubernetes_config = common::kubernetes_config_path(
+        "/tmp",
+        &environment_check.organization_id.as_str(),
+        "dmubm9agk7sr8a8r",
```
deimosfr, line 82: Use var instead

deimosfr (approved):

PR #19, opened 2020-11-06
Infrastructure tests

maathor:


lib/aws/bootstrap/tf-default-vars.j2.tf
```diff
@@ -78,7 +78,7 @@ variable "eks_cluster_name" {
 
 variable "eks_access_cidr_blocks" {
   description = "Kubernetes cluster name"
-  default     = "{{ eks_access_cidr_blocks }}"
+  default     = {{ eks_access_cidr_blocks }}
```
evoxmusic, line 81: @deimosfr no needs to double-quote this?

deimosfr: Nope, because it's a list of string which is expected

PR #26, opened 2020-11-10
add failover test

maathor:


src/transaction.rs
```diff
@@ -497,6 +497,37 @@ impl<'a> Transaction<'a> {
 
                     let applications = apps_result.ok().unwrap();
                     applications_by_environment.insert(target_environment, applications);
+
+                    // build ad well the failover environment, retention could remove the application image
```
deimosfr, line 501: typo ad -> as

test_utilities/src/aws.rs
```diff
@@ -713,3 +713,60 @@ pub fn non_working_environment(context: &Context) -> Environment {
 
     environment
 }
+
+
+pub fn echo_app_environment(context: &Context) -> Environment {
```
deimosfr, line 718: can you please add comment or rename the function, we should understand what role it this

deimosfr (approved):

PR #32, opened 2020-11-14
fix: helm-freeze config issues

deimosfr:


lib/common/chart_values/postgresql/postgresql-values.j2.yaml
```diff
@@ -2,10 +2,6 @@
 ## Please, note that this will override the image parameters, including dependencies, configured to use the global value
 ## Current available global Docker image parameters: imageRegistry and imagePullSecrets
 ##
-
-nameOverride: {{ name }}
```
deimosfr, line 6: To check

lib/common/chart_values/postgresql/postgresql-values.j2.yaml
```diff
@@ -319,12 +379,13 @@ master:
 
   ## Override the service configuration for master
   ##
-  service: {}
+  service:
```
deimosfr, line 382: validate behavior

PR #34, opened 2020-11-16
retry terraform init 5 times, and apply before destroy on terraform

maathor:


src/cmd/terraform.rs
```diff
@@ -13,22 +13,30 @@ use serde_json::Value;
...
-    root_dir: &str,
-    first_time_init_terraform: bool,
-) -> Result<(), SimpleError> {
+fn terraform_exec_with_init_validate(root_dir: &str) -> Result<(), SimpleError> {
     // terraform init
-    let init_args = if first_time_init_terraform {
-        vec!["init"]
-    } else {
-        vec!["init"]
+    let result = retry::retry(Fibonacci::from_millis(3000).take(5), || {
```
deimosfr, line 20: please do not use fibonacci but regular retry instead. No need to wait a long time

deimosfr (approved):

src/cmd/terraform.rs
maathor: k

PR #40, opened 2020-11-20
Feat/redis support

deimosfr:


lib/aws/bootstrap/elasticcache.tf
```diff
@@ -0,0 +1,79 @@
...
+resource "aws_elasticache_subnet_group" "elasticache" {
+  description = "Elasticache linked to ${var.eks_cluster_id}"
+  name = "elasticache-${aws_vpc.eks.id}"
+  subnet_ids = flatten([aws_subnet.elasticache_zone_a.*.id, aws_subnet.elasticache_zone_b.*.id, aws_subnet.elasticache_zone_c.*.id])
+}
+
+# Todo: create a bastion to avoid this
+
+resource "aws_security_group_rule" "elasticache_remote_access" {
+  cidr_blocks       = ["0.0.0.0/0"]
```
maathor, line 72: shluld be a variable sent by the core, maybe this CIDR_BLOCK could be a VPN NAT Addr for some users ? and 0.0.0.0 by default ?

tests/aws/aws_databases.rs
```diff
@@ -522,7 +522,7 @@ fn deploy_a_working_environment_with_mongodb_v4_4() {
 /// test mongodb v3.6 with production environment (DocumentDB)
 #[test]
 #[ignore]
-fn deploy_a_working_production_environment_mongodb_v3_6() {
+fn deploy_a_working_environment_with_production_mongodb_v3_6() {
```
maathor, line 525: we should define a right syntax for this I saw:

mysql_deploy_a_working_environment,
mongodb_v4_4_deploy_a_working_environment,
deploy_a_working_environment_with_production_mongodb_v3_6

I suggest:
[staging/production]_env_[database_type]_[version?] -> with optional version fields

tests/aws/aws_databases.rs
```diff
@@ -714,3 +714,133 @@ fn deploy_a_working_production_environment_with_mysql() {
...
     };
 }
+
+/**
+**
+** Redis tests
+**
+**/
+
+fn test_redis_configuration(context: Context, mut environment: Environment, version: &str) {
```
maathor, line 724: same as before, now we have test suffix

tests/aws/aws_databases.rs
```diff
@@ -714,3 +714,133 @@ fn deploy_a_working_production_environment_with_mysql() {
...
+}
+
+#[test]
+fn redis_v6_deploy_a_working_environment() {
+    let context = context();
+    let mut environment = test_utilities::aws::working_minimal_environment(&context);
+    test_redis_configuration(context, environment, "6.0");
+}
+
+// test Redis v3.6 with production environment (Elasticcache)
```
maathor, line 836: don't see any warning removed :)

lib/aws/bootstrap/elasticcache.tf
deimosfr: Bastion will resolve this, not urgent at the moment

tests/aws/aws_databases.rs
deimosfr: Not a test, just a function

tests/aws/aws_databases.rs
deimosfr: You're right

PR #47, opened 2020-11-25
change logger

maathor:


src/cmd/kubectl.rs
```diff
@@ -514,9 +514,7 @@ where
             ),
         },
         Err(e) => warn!(
-            "Unable to execute describe on secrets: {}. it may not exist anymore?",
```
deimosfr, line 517: Why did you remove this? Wasn't it useful?

tests/aws/aws_databases.rs
```diff
@@ -56,6 +65,12 @@ fn deploy_an_environment_with_3_databases_and_3_apps() {
 fn postgresql_deploy_a_working_development_environment_with_all_options() {
     init();
 
+    let span = span!(
```
deimosfr, line 68: Couldn't it be simplified? Using a function for example?

maathor: unfortunately, span lives on function they are called

src/cmd/kubectl.rs
maathor: nop, IMO

deimosfr (approved):

PR #49, opened 2020-11-27
Feat/redis

deimosfr:


src/cloud_provider/aws/databases/redis.rs
```diff
@@ -577,3 +569,25 @@ impl Backup for Redis {
         unimplemented!()
     }
 }
+
+fn get_redis_version(
+    requested_version: &str,
+    is_managed_service: bool,
+) -> Result<String, StringError> {
+    let mut supported_redis_versions = HashMap::with_capacity(2);
+
```
evoxmusic, line 578: I think you should check the input. The version must be 5 or 6 in this case. Other versions should throw an error.

src/cloud_provider/aws/databases/utilities.rs
```diff
@@ -74,3 +83,82 @@ pub fn delete_terraform_tfstate_secret(
         }
     }
 }
+
+pub fn get_version_semver(version: &str) -> Result<VersionsNumber, StringError> {
```
evoxmusic, line 87: 1/ I'd export this function outside of the Redis struct. 
2/ I'd rename it to `get_semver` or `get_version_number`. `version: &str` already gives details on the expected input.

src/cloud_provider/aws/databases/utilities.rs
```diff
@@ -74,3 +83,82 @@ pub fn delete_terraform_tfstate_secret(
...
+
+    Ok(VersionsNumber {
+        major,
+        minor,
+        patch,
+    })
+}
+
+pub fn get_supported_version_to_use(
+    all_supported_versions: HashMap<&str, &str>,
```
evoxmusic, line 117: I'd suggest using type alias for hashmap key and value, then we can understand what types are expected?

evoxmusic (commented):
Here are my suggestions, and you're probably better to add more tests on the provided version inputs.

src/cloud_provider/aws/databases/redis.rs
deimosfr: It's done by the function called get_supported_version_to_use in this function. If it doesn't match, I throws an error saying:
this {} version is not supported

src/cloud_provider/aws/databases/utilities.rs
deimosfr: done

src/cloud_provider/aws/databases/utilities.rs
evoxmusic: @deimosfr ->

```rust
type ProvidedVersion<'a> = &'a str;
type RealVersion<'a> = &'a str;

//....
pub fn get_supported_version_to_use(
all_supported_versions: HashMap<ProvidedVersion<'a>, RealVersion<'a>>,...
)
```

PR #59, opened 2020-12-02
remove field on docr and force recreate on do

maathor:


lib/digitalocean/bootstrap/doks_gen_kubectl-config.j2.tf
```diff
@@ -17,4 +21,8 @@ resource "digitalocean_spaces_bucket_object" "upload_kubeconfig" {
   key          = "${var.oks_cluster_id}.yaml"
   source       = local_file.kubeconfig.filename
   depends_on = [local_file.kubeconfig, digitalocean_spaces_bucket.space_bucket_kubeconfig]
+
```
deimosfr, line 24: Why is this useful? It's a local file

src/container_registry/docr.rs
```diff
@@ -56,10 +54,9 @@ impl DOCR {
 
     pub fn create_repository(&self, _image: &Image) -> Result<(), EngineError> {
         let mut headers = get_header_with_bearer(&self.api_key);
-        //TODO: receive subscription_tier_slug from core
         let repo = DO_API_Create_repository {
             name: self.registry_name.clone(),
-            subscription_tier_slug: "basic".to_owned(),
+            subscription_tier_slug: "professional".to_owned(),
```
deimosfr, line 59: What is this?

maathor: ok i added some comment ;)

lib/digitalocean/bootstrap/doks_gen_kubectl-config.j2.tf
maathor: you're right

deimosfr (approved):

PR #61, opened 2020-12-04
Digitalocean routers on track

maathor:


lib/digitalocean/charts/nginx-ingress/.helmignore
```diff
@@ -0,0 +1,21 @@
+# Patterns to ignore when building packages.
```
deimosfr, line 1: Why did you imported it inside DO? It's available in the common charts

src/cloud_provider/digitalocean/application.rs
```diff
@@ -11,14 +11,16 @@ use crate::cloud_provider::service::{
     StatelessService,
 };
 use crate::cloud_provider::DeploymentTarget;
+use crate::cmd::helm::Timeout;
 use crate::cmd::structs::HelmHistoryRow;
```
deimosfr, line 15: can you please remove unused imports as GitHub noticed

src/cloud_provider/digitalocean/common.rs
```diff
@@ -23,21 +28,41 @@ pub fn kubernetes_config_path(
...
-    let kubeconfig = download_space_object(
-        spaces_access_id,
-        spaces_secret_key,
-        kubernetes_config_bucket_name.as_str(),
-        kubernetes_config_object_key.as_str(),
-        region,
-    );
-    match kubeconfig {
-        Ok(body) => {
+    let result = retry::retry(Fixed::from_millis(3000).take(5), || {
```
deimosfr, line 31: shouldn't we add a sleep between each try?

src/cloud_provider/digitalocean/common.rs
```diff
@@ -23,21 +28,41 @@ pub fn kubernetes_config_path(
...
+        }
+    });
+
+    match result {
+        Ok(downloaded) => {
             let mut file =
                 File::create(kubernetes_config_file_path.clone()).expect("unable to create file");
-            file.write_all(body.as_bytes()).expect("unable to write");
+            file.write_all(downloaded.as_bytes())
+                .expect("unable to write");
```
deimosfr, line 56: can you please give more context on the message, it would be easier for debug purpose

lib/digitalocean/bootstrap/tf-default-vars.j2.tf
```diff
@@ -4,12 +4,7 @@ variable "digitalocean_region" {
   type    = string
 }
 
-// container registry
-variable "container_registry_name" {
-  default = "qovery"
-  type = string
-}
-
+// kubernetes MASTER cluster
```
deimosfr, line 7: duplicate line

src/cloud_provider/digitalocean/common.rs
```diff
@@ -54,7 +79,11 @@ struct Cluster {
 }
 */
 
-pub fn get_uuid_of_cluster(token: &str, kubeID: &str) -> Result<String, SimpleError> {
+// retreive the digital ocean uuid of the kube cluster from our cluster name
+pub fn get_uuid_of_cluster_from_name(
```
deimosfr, line 83: I can see in the code a comment like "retrieve the cluster uuid, useful to link DO registry to k8s cluster". Can you please add some comments to explain why we need to have it. Isn't it possible to use the registry without the ID of the container?

src/cloud_provider/digitalocean/kubernetes/mod.rs
```diff
@@ -1,5 +1,5 @@
 use crate::cloud_provider::common::workerNodeDataTemplate::WorkerNodeDataTemplate;
-use crate::cloud_provider::digitalocean::common::get_uuid_of_cluster;
+use crate::cloud_provider::digitalocean::common::get_uuid_of_cluster_from_name;
```
deimosfr, line 2: please also clean non used imported warning

src/cloud_provider/digitalocean/kubernetes/mod.rs
```diff
@@ -85,8 +85,8 @@ impl<'a> DOKS<'a> {
         let mut context = TeraContext::new();
 
         // VPC segmentation
-        let vpc_name = "qovery";
-        context.insert("vpc_name", vpc_name);
+        let vpc_name = format!("qovery-{}", &self.id);
```
deimosfr, line 88: Warning, I've already fixed this yesterday. I think you'll have to rebase. Also, I think prefixing is a good idea (I didn't did it) https://github.com/Qovery/engine/pull/60

src/cloud_provider/digitalocean/router.rs
```diff
@@ -0,0 +1,494 @@
...
+            crate::crypto::to_sha1_truncate_16(self.default_domain.as_str());
+
+        context.insert("router_default_domain", self.default_domain.as_str());
+        context.insert(
+            "router_default_domain_hash",
+            router_default_domain_hash.as_str(),
+        );
+        context.insert("custom_domains", &custom_domain_data_templates);
+        context.insert("routes", &route_data_templates);
+        context.insert("spec_acme_email", "tls@qovery.com"); // TODO CHANGE ME
```
deimosfr, line 216: please update with the correct changeme

src/cloud_provider/digitalocean/router.rs
```diff
@@ -0,0 +1,494 @@
...
+                            error!(
+                                "Waiting Digital Ocean LoadBalancer endpoint to be available to be able to configure TLS"
+                            );
+                            OperationResult::Retry(err)
+                        }
+                    }
+                },
+            );
+
+            match external_ingress_hostname_custom_result {
```
deimosfr, line 430: is nlb the name of DO for LBs? If no please update to keep the correct naming convention

src/cloud_provider/digitalocean/common.rs
maathor: nop

maathor: it's due a call that never give the control back ... will work on it on the futur PR.

src/cloud_provider/digitalocean/common.rs
maathor: done add some comment upper ;)

src/cloud_provider/digitalocean/router.rs
maathor: done

lib/digitalocean/bootstrap/tf-default-vars.j2.tf
```diff
@@ -4,12 +4,6 @@ variable "digitalocean_region" {
   type    = string
 }
 
-// container registry
-variable "container_registry_name" {
```
deimosfr, line 8: no need it? or as a variable?

lib/digitalocean/bootstrap/tf-default-vars.j2.tf
```diff
@@ -37,7 +31,7 @@ variable "oks_master_size" {
 }
 
 variable "oks_master_node_count" {
-  default = 2
+  default = 4
```
deimosfr, line 34: oO? Why do we need to setup master nodes? 3 is better than 4. Please choose between 3 or 5 for the quorum vote

src/cloud_provider/digitalocean/kubernetes/mod.rs
```diff
@@ -172,7 +170,7 @@ impl<'a> DOKS<'a> {
 
         context.insert("oks_version", &self.version());
 
-        context.insert("oks_master_size", "s-1vcpu-2gb");
+        context.insert("oks_master_size", "s-2vcpu-4gb");
```
deimosfr, line 173: Shouldn't we add variable here?

test_utilities/src/aws.rs
```diff
@@ -800,3 +800,53 @@ pub fn environment_only_http_server(context: &Context) -> Environment {
     }
 }
 
+pub fn environment_only_http_server_router(context: &Context) -> Environment {
```
deimosfr, line 803: What is this? If this is the "minimum", shouldn't we just update the current existing one instead?

deimosfr (approved):

PR #84, opened 2020-12-11
Feat do add tags and firewall

maathor:


lib/digitalocean/bootstrap/oks-firewall.tf
```diff
@@ -0,0 +1,22 @@
+# due to tags, (applied on kube cluster and pools), all droplets (scaled) will me automatically attached to the firewall below
+resource "digitalocean_tag" "cluster_tag" {
+  name = var.oks_master_name
```
deimosfr, line 3: can you please same tags we've set for AWS as well

deimosfr: I also advise you to look on how it has been done on AWS

maathor: tags doesn't works with the same behavior of aws
They are very linked with do firewalls

deimosfr (approved):

PR #85, opened 2020-12-11
chart engine: use do instead digitalocean

maathor:


deimosfr:
why?

maathor:
> why?

to be consistent with core, we use [aws,gcp,...]

deimosfr (approved):

PR #92, opened 2020-12-14
fix-fields-when-calling-kubectl-get-svc-ingress-router

maathor:


src/object_storage/do_space.rs
```diff
@@ -44,7 +44,7 @@ pub(crate) async fn download_space_object(
                     Ok(_) => info!("File {} is well downloaded", path_to_download),
                     Err(e) => error!("{:?}", e),
                 },
-                Err(e) => error!("Unable to create file"),
+                Err(e) => error!("Unable to create file, maybe this file already exists..."),
```
deimosfr, line 47: can you please be more explicit on the context. like adding file name, why etc..

maathor: this fn is refactored,  now we only download it if file does't exist

deimosfr (approved):

PR #103, opened 2020-12-21
check domain propagation

evoxmusic:


src/cloud_provider/aws/router.rs
```diff
@@ -282,34 +288,98 @@ impl Service for Router {
...
+                self.name_with_id()
+            )),
+            self.context.execution_id(),
+        ));
+
+        let mut resolver_options = ResolverOpts::default();
+        resolver_options.cache_size = 0;
+        resolver_options.use_hosts_file = false;
+
+        let resolver = match Resolver::new(ResolverConfig::google(), resolver_options) {
```
deimosfr, line 309: Please use  cloudflare() instead. It's more relevant due to our config

evoxmusic: Checking through google is maybe a better idea due to the fact most users/isp may use Google resolver instead of Cloudflare. Don't you think?

deimosfr: You're right, but it may take more time before being available. Let's see if we encounter timeouts. If it's the case, we'll leverage the retry time or switch to cloudflare.

PR #123, opened 2021-01-12
send task progress on long deployment

evoxmusic:


src/cloud_provider/service.rs
```diff
@@ -1161,6 +1168,98 @@ pub fn do_stateless_service_cleanup(
     Ok(())
 }
 
+/// This function call (start|pause|delete)_in_progress function every 10 seconds when a
+/// long blocking task is running.
+pub fn send_progress_on_long_task<S, R, F>(service: &S, action: Action, long_task: F) -> R
```
evoxmusic, line 1173: the most important part to check is here

src/cloud_provider/service.rs
```diff
@@ -1161,6 +1168,98 @@ pub fn do_stateless_service_cleanup(
...
+    F: Fn() -> R,
+{
+    let listeners = std::clone::Clone::clone(service.listeners());
+
+    let progress_info = ProgressInfo::new(
+        service.progress_scope(),
+        Info,
+        match waiting_message {
+            Some(message) => Some(message.into()),
+            _ => None,
```
deimosfr, line 1220: Waiting message should not be an option, otherwise, I don't get the interest in it

evoxmusic: We don't necessarily show the progress to the user, this function can be used for anything, and the progress can be used to know if the task is really running or not.

deimosfr (approved):

PR #128, opened 2021-01-15
add ObjectStorage.put(..) function to upload object on S3 and Spaces

evoxmusic:
…paces (not implemented)

### DIFF
--- a/src/object_storage/s3.rs
+++ b/src/object_storage/s3.rs
@@ -159,4 +159,21 @@ impl ObjectStorage for S3 {
             Err(err) => Err(self.engine_error(EngineErrorCause::Internal, format!("{:?}", err))),
         }
     }
+
+    fn put(&self, bucket_name: &str, object_key: &str, file_path: &str) -> Result<(), EngineError> {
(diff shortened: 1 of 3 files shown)


src/object_storage/s3.rs
```diff
@@ -159,4 +159,21 @@ impl ObjectStorage for S3 {
             Err(err) => Err(self.engine_error(EngineErrorCause::Internal, format!("{:?}", err))),
         }
     }
+
+    fn put(&self, bucket_name: &str, object_key: &str, file_path: &str) -> Result<(), EngineError> {
+        cast_simple_error_to_engine_error(
+            self.engine_error_scope(),
+            self.context().execution_id(),
+            crate::cmd::utilities::exec_with_envs(
```
deimosfr, line 167: why don't you use rusoto for it? This time it can be async, we don't care :)

evoxmusic: if we use Rusoto here, then I suggest to change it everywhere to be consistent.

deimosfr (approved):

PR #138, opened 2021-01-25
feat: temporary remove kube-state-metrics

deimosfr:
Before, we were using kube-state-metrics from Prometheus operator chart. But because we can't control resource usage, it may be preferable to use the dedicated chart for it.
I disable it because Terraform was failing on chart upgrade (because of pod CrashloopBackoff)


erebe:
:+1:

erebe (approved):

PR #140, opened 2021-01-26
DNS: Wait for CNAME to be valid in case of custom domains

erebe:
When we have a custom domain, the user needs to create a CNAME toward qovery.
This PR check and wait for the CNAME to be valid before trying to deploy the router


src/cloud_provider/utilities.rs
```diff
@@ -242,33 +244,103 @@ fn get_version_number(version: &str) -> Result<VersionsNumber, StringError> {
...
+    //  We want to avoid cache and using hostfile of the host, as some provider force caching
+    //  which lead to stale response
+    resolver_options.cache_size = 0;
+    resolver_options.use_hosts_file = false;
+
+    Resolver::new(ResolverConfig::cloudflare(), resolver_options)
+        .expect("Invalid cloudflare DNS resolver configuration")
+}
+
+fn get_domain_under_cname(resolver: &Resolver, cname: &str) -> Option<String> {
```
deimosfr, line 259: no clear, I suggest to rename to get_cname_record_value

tests/dns.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[test]
+fn test_bad_add() {
+    use std::net::*;
+    use trust_dns_resolver::Resolver;
+    use trust_dns_resolver::config::*;
+
+// Construct a new Resolver with default configuration options
+    let resolver = Resolver::new(ResolverConfig::default(), ResolverOpts::default()).unwrap();
+    let ret = resolver.lookup("www.fillbit.io.", RecordType::CNAME).unwrap();
```
deimosfr, line 11: please use another random name, one of qovery for exmaple

src/cloud_provider/utilities.rs
erebe: done

tests/dns.rs
erebe: Was not supposed to commit it.
But moved the test and added a cname on qovery.io for that

deimosfr (approved):

PR #181, opened 2021-03-22
Use sanitized name to get events for a pod

erebe:


src/cmd/kubectl.rs
```diff
@@ -728,17 +728,27 @@ where
     )
 }
 
-pub fn kubectl_exec_get_event<P>(
+pub fn kubectl_exec_get_event_for_pod<P>(
```
deimosfr, line 731: what about other services you don't have the pod name? Example: statefulset running MySQL/Postgres...

PR #186, opened 2021-03-25
feat: add docker network support for CI

deimosfr:


src/build_platform/local_docker.rs
```diff
@@ -74,6 +74,10 @@ impl LocalDocker {
             vec!["build"]
         };
 
+        let args = self.context.docker_build_options();
+        args.iter()
```
erebe, line 78: maybe more readable if you turn it into a for loop.
for v in args.iter() {
   for string in v.iter() {
       docker_args.push(String::as_str(string))
    }
}

test_utilities/src/utilities.rs
```diff
@@ -26,6 +26,20 @@ use serde::{Deserialize, Serialize};
...
 
+pub fn context() -> Context {
+    let execution_id = execution_id();
+    let home_dir = std::env::var("WORKSPACE_ROOT_DIR").unwrap_or(home_dir().unwrap().to_str().unwrap().to_string());
+    let lib_root_dir = std::env::var("LIB_ROOT_DIR").expect("LIB_ROOT_DIR is mandatory");
+
+    let metadata = Metadata {
+        dry_run_deploy: Option::from(false),
+        resource_expiration_in_seconds: Some(3600),
+        docker_build_options: Some("--network 'host'".to_string()),
```
erebe, line 37: Maybe it can be good to add as a default settings even in prod ?
We are doing also dind so it can help 
What do you think ?

deimosfr: At the moment there is no need in prod for this, only for tests purpose. May be later when the need will be there

erebe (approved):

PR #194, opened 2021-04-01
feat: adding default cors support #DEV-824

deimosfr:


erebe:
Put a comment (with ticket link) to say that this should be fixed in the future, as it allows every orgin at the moment

deimosfr:
> Put a comment (with ticket link) to say that this should be fixed in the future, as it allows every origin at the moment

I fixed definitively the issue

PR #197, opened 2021-04-05
Feat/infra pause

deimosfr:
This PR enables infra pause for AWS:
* Pause will destroy nodes group
* Restore will re-run deploy to get everything deployed
* Updated pod policy to critical elements to ensure they will be correctly restored and prioritized
* Added prometheus adapter to support custom metrics and report the to k8s API
* Checking no jobs are running on the engine before starting pause using custom metrics api
* Refactored terraform force upgrade hack to speed up re-deployments


erebe:
Can you describe a bit what this does ? and how it is suppose to do it to achieve it ?

deimosfr:
> Can you describe a bit what this does ? and how it is suppose to do it to achieve it ?

Read the code! #Toxic

Where do you want to see this, in the code? In the PR description? Somewhere else?

erebe:
In the PR is fine ;)

erebe (approved):

PR #198, opened 2021-04-06
Do not abort at first error when retrieving debug info

erebe:
- Try to fetch as much as possible info instead of aborting
    at first error

### DIFF
--- a/src/cloud_provider/service.rs
+++ b/src/cloud_provider/service.rs
@@ -944,7 +944,8 @@ where
             selector.as_str(),
             kubernetes.cloud_provider().credentials_environment_variables(),
         ),
-    )?;
+    )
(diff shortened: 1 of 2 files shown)


src/cloud_provider/service.rs
```diff
@@ -944,7 +944,8 @@ where
             selector.as_str(),
             kubernetes.cloud_provider().credentials_environment_variables(),
         ),
-    )?;
+    )
+    .unwrap_or_else(|_| vec!["No Logs".to_string()]);
```
deimosfr, line 948: Can you create a longer message plz, easier to find in the logs and more understandable (add context or explanation). Thx

erebe: better ?

deimosfr (approved):

PR #200, opened 2021-04-14
feat: reporting terraform error logs to user

deimosfr:


src/cloud_provider/service.rs
```diff
@@ -539,14 +539,38 @@ where
...
+            return match crate::cmd::terraform::terraform_init_validate_plan_apply(
+                workspace_dir.as_str(),
+                service.context().is_dry_run_deploy(),
+            ) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    // join error messages and Terraform error on the same string
+                    let mut error_complete = "".to_string();
+
+                    if e.message.is_some() {
```
evoxmusic, line 551: I would recommend using the following syntax instead

```rust
if let Some(message) = e.message {
  error_complete.push_str(message.as_str());
}
```

src/cloud_provider/service.rs
```diff
@@ -539,14 +539,38 @@ where
...
+                        error_complete.push_str(e.message.unwrap().as_str());
+                    }
+
+                    if e.logs.is_some() {
```
evoxmusic, line 555: same here

PR #210, opened 2021-04-22
chore: adding digital mobius

MacLikorne:


lib/common/bootstrap/charts/digital-mobius/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+podAnnotations: {}
+
+podSecurityContext: {}
+
+securityContext: {}
+
+resources:
+  limits:
+    cpu: 100m
+    memory: 128Mi
```
deimosfr, line 42: can you please reduce it to 32Mib

lib/common/bootstrap/charts/digital-mobius/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+  requests:
+    cpu: 100m
+    memory: 128Mi
```
deimosfr, line 45: this one as well

deimosfr (approved):

deimosfr (approved):

PR #220, opened 2021-05-05
Feat/add more logs to docker push

deimosfr:


src/container_registry/utilities.rs
```diff
@@ -37,8 +38,22 @@ pub fn docker_tag_and_push_image(
...
+        || match cmd::utilities::exec_with_envs_and_output(
+            "docker",
+            vec!["push", dest.as_str()],
+            docker_envs.clone(),
+            |line| {
+                let line_string = line.unwrap();
+                info!("{}", line_string.as_str());
+            },
+            |line| {
+                let line_string = line.unwrap();
```
erebe, line 52: do a `line.map(|| info())` instead of unwraping

PR #223, opened 2021-05-06
fix: add new prune commands

deimosfr:
In order to maximize the space gain after a docker prune, I've added volumes and containers

### DIFF
--- a/src/build_platform/local_docker.rs
+++ b/src/build_platform/local_docker.rs
@@ -518,52 +518,33 @@ fn check_docker_space_usage_and_clean(
     ))
 }
 
-fn docker_prune_images(envs: Vec<(&str, &str)>) -> Result<Vec<String>, SimpleError> {
-    let docker_image_prune_args = vec!["image", "prune", "-a", "-f"];
-    let docker_builder_prune_args = vec!["builder", "prune", "-a", "-f"];
+fn docker_prune_images(envs: Vec<(&str, &str)>) -> Result<(), SimpleError> {
+    let all_prunes_commands = vec![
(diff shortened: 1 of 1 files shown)


src/build_platform/local_docker.rs
```diff
@@ -518,52 +518,33 @@ fn check_docker_space_usage_and_clean(
...
+        vec!["volume", "prune", "-f"],
+    ];
+
+    for prune in all_prunes_commands {
+        match cmd::utilities::exec_with_envs_and_output(
+            "docker",
+            prune,
+            envs.clone(),
+            |line| {
+                let line_string = line.unwrap();
```
erebe, line 535: line.unwrap again :o: 
use `unwrap_or_default` or `map`

src/build_platform/local_docker.rs
```diff
@@ -518,52 +518,33 @@ fn check_docker_space_usage_and_clean(
...
-        Ok(mut x) => {
-            let mut image_prune_lines = result_docker_image_prune.unwrap();
-            image_prune_lines.append(&mut x);
-            let all_lines = image_prune_lines;
-            Ok(all_lines)
-        }
-        Err(x) => Err(x),
+fn docker_prune_images(envs: Vec<(&str, &str)>) -> Result<(), SimpleError> {
+    let all_prunes_commands = vec![
+        vec!["container", "prune", "-f"],
```
erebe, line 523: Why not ` docker system prune -a -f` ?

deimosfr: Because it removes networks which breaks docker then if you don't restart the daemon, which is not possible because of PID 1

PR #231, opened 2021-05-10
feat: support AWS Kubernetes cluster upgrade

deimosfr:


src/cloud_provider/utilities.rs
```diff
@@ -218,22 +220,47 @@ pub fn generate_supported_version(
...
 pub struct VersionsNumber {
     pub(crate) major: String,
     pub(crate) minor: Option<String>,
     pub(crate) patch: Option<String>,
 }
 
+impl fmt::Display for VersionsNumber {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut version = vec![];
+        version.push(format!("{}", &self.major));
```
evoxmusic, line 233: you can replace with `let mut version = vec![self.major.clone()];`

evoxmusic: Plus, `version.push(format!("{}", &self.major));` should be `version.push(self.major.clone());`

src/cloud_provider/utilities.rs
```diff
@@ -218,22 +220,47 @@ pub fn generate_supported_version(
...
+        if self.minor.is_some() {
+            version.push(format!("{}", &self.minor.clone().unwrap()))
```
evoxmusic, line 235: same here. Why do you use `format!("{}"...)` instead of directly cloning with `self.minor.clone()`.
Plus, you can remove `&self` as it's already a reference. Cf. method signature with `&self` `fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {`

src/cloud_provider/utilities.rs
```diff
@@ -218,22 +220,47 @@ pub fn generate_supported_version(
...
+        }
+        if self.patch.is_some() {
+            version.push(format!("{}", &self.patch.clone().unwrap()))
```
evoxmusic, line 238: same here

tests/aws/aws_kubernetes.rs
```diff
@@ -64,6 +71,82 @@ fn create_and_destroy_eks_cluster(region: &str, secrets: FuncTestsSecrets, test_
...
+            region,
+            &aws,
+            &cloudflare,
+            eks_options(secrets.clone()),
+            nodes.clone(),
+        );
+
+        // Deploy
+        if let Err(err) = tx.create_kubernetes(&kubernetes) {
+            panic!("{:?}", err)
```
evoxmusic, line 84: Ouch!!! Don't use `panic!(..)` this is going to crash the application without any recovery. This is super dangerous ☠️. Return an error instead. You will avoir to spend your weekend to troubleshoot why the engine crash :)

tests/aws/aws_kubernetes.rs
```diff
@@ -64,6 +71,82 @@ fn create_and_destroy_eks_cluster(region: &str, secrets: FuncTestsSecrets, test_
...
+            generate_cluster_id(region).as_str(),
+            upgrade_to_version,
+            region,
+            &aws,
+            &cloudflare,
+            eks_options(secrets),
+            nodes,
+        );
+        if let Err(err) = tx.create_kubernetes(&kubernetes) {
+            panic!("{:?}", err)
```
evoxmusic, line 105: same here

tests/aws/aws_kubernetes.rs
```diff
@@ -64,6 +71,82 @@ fn create_and_destroy_eks_cluster(region: &str, secrets: FuncTestsSecrets, test_
...
+            TransactionResult::Ok => assert!(true),
+            TransactionResult::Rollback(_) => assert!(false),
+            TransactionResult::UnrecoverableError(_, _) => assert!(false),
+        };
+
+        // Destroy
+        // There is a bug with the current version of Terraform (0.14.10) where the destroy fails, but it works
+        // It doesn't find any helm charts after destroying the workers and charts have already been destroyed
+        if let Err(err) = tx.delete_kubernetes(&kubernetes) {
+            panic!("{:?}", err)
```
evoxmusic, line 117: same here

tests/aws/aws_kubernetes.rs
evoxmusic: ha sorry it's tests :) so it's good

src/cloud_provider/utilities.rs
deimosfr: Why is it better to do version.push(self.major.clone()); ?

evoxmusic: because the format is unnecessary here. `format!(..)` is good to use when you want to concatenate strings/literals. What you want here is simply put your `self.major` string into a vec. You only need to clone it and put it directly into it.

deimosfr: By the way, I just seen that Clippy suggest to use &self.major.to_string()

evoxmusic: `self.major.clone()` is good as well ^^

src/cloud_provider/utilities.rs
```diff
@@ -218,22 +220,46 @@ pub fn generate_supported_version(
 
 // unfortunately some proposed versions are not SemVer like Elasticache (6.x)
 // this is why we need ot have our own structure
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
```
erebe, line 223: Where do you use PartialEq ?

erebe (approved):

src/cloud_provider/utilities.rs
deimosfr: Because of tests

PR #233, opened 2021-05-14
release: pleco v0.7.17 & digital mobius v0.1.4

MacLikorne:


lib/aws/bootstrap/tf-providers-aws.j2.tf
```diff
@@ -14,7 +14,7 @@ terraform {
     }
     helm = {
       source = "hashicorp/helm"
-      version = "~> 1.3.2"
+      version = "~> 2.1.2"
```
deimosfr, line 17: please remove this as we're going to remove it shortly

MacLikorne: done

PR #234, opened 2021-05-16
feat: reduced resource usage for custom domains

deimosfr:
* For production usage, autoscaler is enabled with more resources
* For dev usage, autoscaler is disabled, only 1 nginx is running with a small amount of resources


evoxmusic (approved):
LGTM

PR #238, opened 2021-05-19
fix: decrease min available budget

MacLikorne:


MacLikorne:
Maybe don't put it hard coded but link to replicas count.

deimosfr (approved):

PR #240, opened 2021-05-26
fix: ec2 key pairs tagging

MacLikorne:


lib/aws/bootstrap/ec2-ssh-keys.j2.tf
```diff
@@ -3,5 +3,5 @@ resource "aws_key_pair" "qovery_ssh_key_{{ kubernetes_cluster_id }}" {
   public_key = var.ec2_ssh_default_key.public_key
 
   # Don't add tags, there is an issue, the key will disappear, making the autoscaler stuck because of this (in the best case)
```
deimosfr, line 5: no need this comment anymore

lib/aws/bootstrap/eks-master-cluster.j2.tf
```diff
@@ -9,6 +9,7 @@ locals {
     ClusterId = var.kubernetes_cluster_id,
     ClusterName = var.kubernetes_cluster_name,
     Region = var.region
+    CreationDate = var.kubernetes_cluster_creation_date
```
deimosfr, line 12: can't we just add it where it's necessary?

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -320,6 +321,7 @@ impl<'a> EKS<'a> {
         // AWS - EKS
         context.insert("eks_cidr_subnet", &eks_cidr_subnet.clone());
         context.insert("kubernetes_cluster_name", &self.name());
+        context.insert("kubernetes_cluster_creation_date", &Utc::now().format("%Y-%m-%d %H:%M:%S").to_string());
```
deimosfr, line 324: It won't work as the date will be reset anytime the engine will be triggered

deimosfr:
Please wait tests to pass before merging it

lib/aws/bootstrap/eks-master-cluster.j2.tf
```diff
@@ -6,13 +6,18 @@ locals {
 
 locals {
   tags_eks = {
-    ClusterId = var.kubernetes_cluster_id,
+    ClusterId = var.kubernetes_cluster_id
     ClusterName = var.kubernetes_cluster_name,
     Region = var.region
+    creationDate = time_static.on_cluster_create.rfc3339
```
deimosfr, line 12: Are you sure it works? Doesn't it change the date on re-apply?

MacLikorne: Tried it, doesn't change things after first apply.

PR #241, opened 2021-05-26
feat: adding helm deployment support for AWS

deimosfr:
Implemented all charts for AWS and logic, not plugged yet


src/cloud_provider/helm.rs
```diff
@@ -0,0 +1,204 @@
...
+use thread::spawn;
+
+#[derive(Clone)]
+pub enum HelmAction {
+    Deploy,
+    Destroy,
+    Skip,
+}
+
+#[derive(Copy, Clone)]
```
erebe, line 17: Adding Debug trait can be useful

src/cloud_provider/helm.rs
```diff
@@ -0,0 +1,204 @@
...
+        HelmChartNamespaces::NginxIngress => "nginx-ingress",
+        HelmChartNamespaces::Qovery => "qovery",
+    }
+    .to_string()
+}
+
+pub trait HelmChart {
+    fn check_prerequisites(&self) -> Result<(), SimpleError> {
+        let chart = self.get_chart_info();
+        for file in chart.values_files.clone() {
```
erebe, line 81: use into_iter, or iter()

src/cloud_provider/helm.rs
deimosfr: I'll see when doing the integration

src/cloud_provider/helm.rs
deimosfr: Using iter() won't let me escape from the lambda on the first error

src/cloud_provider/helm.rs
```diff
@@ -0,0 +1,204 @@
...
+        Ok(())
+    }
+
+    fn get_chart_info(&self) -> &ChartInfo;
+
+    fn namespace(&self) -> String {
+        get_chart_namespace(self.get_chart_info().namespace)
+    }
+
+    fn pre_exec(&self, _kubernetes_config: &Path, _envs: &Vec<(String, String)>) -> Result<(), SimpleError> {
```
erebe, line 104: Change `&Vec<(String, String)>)` to `&[String, String)>]`

erebe (approved):

PR #254, opened 2021-06-07
feat: move out terraform folder form engine archive

benjaminch:
This CL introduces:
- rename generated archive with the proper extension 'tgz' instead of
  'tar.gz'
- a way to exclude files / folders for generated archive
- excludes terraform's folder and hcl

DEV-897

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::fs;
 use std::fs::{create_dir_all, File};
 use std::io::Error;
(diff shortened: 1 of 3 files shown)


src/fs.rs
```diff
@@ -65,16 +66,31 @@ where
...
+        ".terraform.lock.hcl",
+        ".terraform",]
+        .into_iter()
+        .collect();
+
+    for entry in WalkDir::new(workspace_dir.clone())
+        .into_iter()
+        .filter_entry(|e| !excluded_files.contains(e.file_name().to_str().unwrap()))
+    {
+        let entry = entry.unwrap();
```
deimosfr, line 85: Note: I recommend to use expect with a specific message when you can

deimosfr (approved):

src/fs.rs
benjaminch: Done 👍

PR #259, opened 2021-06-09
Add support for mono-repo with multiple build context

erebe:
- Change to support mono-repo, aka being able to build image from other directory than the root of the repository.

- Fix: Don't try to use buildkit if the specified Dockerfile is not present in the repository


deimosfr (approved):

deimosfr:
@benjaminch it concerns what you're working on as well

benjaminch:
Yep, it touches my change as well, I will update on my side :)

benjaminch (approved):

PR #264, opened 2021-06-17
Fix infra do

erebe:


benjaminch (approved):

deimosfr:
It should not be merged, but refactored with correct valued, I'll handle it once I'll start DO tests

PR #269, opened 2021-06-22
Force deployment restart on helm upgrade

erebe:
+ In order to implement restart, we want to force rollout even
    if nothing changes in the deployment
  + CREATE action should be renamed DEPLOY to be more insightfull


deimosfr:
It looks like your forgot statefulsets

erebe:
I 

> It looks like your forgot statefulsets

I thought we don't want to restart databases ?
Statefulset are annoted ```{%- if is_storage %}```

erebe:
When do we use statefulset for applications ?

deimosfr:
when there is a disk attached

deimosfr (approved):

PR #275, opened 2021-06-24
fix: add message when tf config file is not found and can't be rendered

deimosfr:


benjaminch (approved):
remove extra `.` after `?`

PR #276, opened 2021-06-24
feat: start integration scaleway

benjaminch:


lib/scaleway/bootstrap/master-cluster.tf
```diff
@@ -0,0 +1,41 @@
+resource "scaleway_k8s_cluster" "kubernetes_cluster" {
+  name    = "qovery-${var.kubernetes_cluster_id}"
+  version = "1.21.1"
```
deimosfr, line 3: 1. please use 1.20.x
2. Like DO, user should only set the majo+minor version, we handle the rest (update). Be sure we can set only major+minor, otherwise you'll have to handle it in the engine by calling scaleway API

lib/scaleway/bootstrap/master-cluster.tf
```diff
@@ -0,0 +1,41 @@
...
+  cni     = "cilium"
+
+  tags    = ["qovery", "integration-test"] # TODO(benjaminch): put more usefull data in tags
```
deimosfr, line 6: Please use the same tags than on AWS

lib/scaleway/bootstrap/master-cluster.tf
```diff
@@ -0,0 +1,41 @@
...
+
+  autoscaler_config {
+    disable_scale_down              = false
+    scale_down_delay_after_add      = "5m"
```
deimosfr, line 10: use default settings instead plz (10m) as this is the default used by cluster-autoscaler as well

lib/scaleway/bootstrap/master-cluster.tf
```diff
@@ -0,0 +1,41 @@
...
+    estimator                       = "binpacking"
+    expander                        = "random"
+    ignore_daemonsets_utilization   = true
+    balance_similar_node_groups     = true
+    expendable_pods_priority_cutoff = -5
+  }
+}
```
deimosfr, line 17: Please add max_graceful_termination_sec and set it to 60m as some pods can have long termination

lib/scaleway/bootstrap/master-cluster.tf
```diff
@@ -0,0 +1,41 @@
...
+  tags    = ["qovery", "integration-test"] # TODO(benjaminch): put more usefull data in tags
+
+  autoscaler_config {
+    disable_scale_down              = false
+    scale_down_delay_after_add      = "5m"
+    estimator                       = "binpacking"
+    expander                        = "random"
+    ignore_daemonsets_utilization   = true
+    balance_similar_node_groups     = true
+    expendable_pods_priority_cutoff = -5
```
deimosfr, line 15: I'm not sure about the value of this setting, we'll have to dig into it

lib/scaleway/bootstrap/tf-default-vars.j2.tf
```diff
@@ -0,0 +1,148 @@
...
+  type        = string
+}
+
+variable "scaleway_ks_version" {
+  description = "Kubernetes cluster version"
+  default = "{{ scaleway_ks_version }}"
+  type    = string
+}
+
+# kubernetes WORKER second cluster
```
deimosfr, line 74: second cluster?

lib/scaleway/bootstrap/master-cluster.tf
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+resource "scaleway_k8s_pool" "kubernetes_cluster_pool" {
+  cluster_id    = scaleway_k8s_cluster.kubernetes_cluster.id
+  name          = "qovery-${var.kubernetes_cluster_id}"
+  node_type     = "DEV1-L"
+
+  # use Scaleway built-in cluster autoscaler
+  autoscaling   = true
```
deimosfr, line 25: you should use your variable scaleway_ks_pool_autoscale

lib/scaleway/bootstrap/tf-default-vars.j2.tf
```diff
@@ -0,0 +1,148 @@
...
+
+variable "scaleway_ks_pool_name" {
```
deimosfr, line 76: warn: we can have several pools of nodes

lib/scaleway/bootstrap/tf-default-vars.j2.tf
```diff
@@ -0,0 +1,148 @@
...
+  type = bool
+}
+
+variable "metrics_history_enabled" {
+  description = "Enable metrics history"
+  default = {{ metrics_history_enabled }}
+  type = bool
+}
+
+# Force helm upgrade
```
deimosfr, line 143: not needed anymore, you can remove  this

lib/scaleway/bootstrap/master-cluster.tf
benjaminch: Unfortunately, Scaleway doesn't seem to support it for the time being, I will ask the question though.

lib/scaleway/bootstrap/master-cluster.tf
benjaminch: Removed it for the time being, to be investigated more in depth.

lib/scaleway/bootstrap/master-cluster.tf
benjaminch: Actually they do support it, that's great :)

lib/scaleway/bootstrap/ks-gen-kubeconfig.tf
```diff
@@ -0,0 +1,6 @@
+resource "local_file" "kubeconfig" {
+  filename = "test-cluster/test-cluster.yaml" # TODO(benjaminch): use "{{ s3_kubeconfig_bucket }}/${var.kubernetes_cluster_id}.yaml"
+  content = scaleway_k8s_cluster.kubernetes_cluster.kubeconfig[0].config_file
+  file_permission = "0644"
+  depends_on = [scaleway_k8s_pool.kubernetes_cluster_worker] # TODO(benjaminch): use `scw_ks_worker_node in scw_ks_worker_nodes` to get all nodes names
```
deimosfr, line 5: don't depends on worker nodes (never ever) as they can change and dependency is hard in TF (if something depends on it and dependency is not there, it's deleted). In addition to this, the kubeconfig depends ONLY on masters and not workers, so you can safely update this dependency to master

lib/scaleway/bootstrap/ks-master-cluster.tf
```diff
@@ -0,0 +1,22 @@
+resource "scaleway_k8s_cluster" "kubernetes_cluster" {
+  name    = "test-cluster" # TODO(benjaminch) : use : "qovery-${var.kubernetes_cluster_id}"
+  version = "1.20"
+  cni     = "cilium"
+
+  tags    =  [for i, v in local.tags_ks : "${i}=${v}"] # NOTE: Scaleway doesn't support KV style tags
```
deimosfr, line 6: You should declare a local variable and set it directly there. Then call this variable on every tags

lib/scaleway/bootstrap/ks-object-bucket.tf
```diff
@@ -0,0 +1,11 @@
+# TODO(benjaminch): To be handled directly in the engine to properly handle reinstall
+resource "scaleway_object_bucket" "object-bucket" {
+  # NOTE: name supports only alpha-numerics!
+  name = "test-cluster" # TODO(benjaminch): use var.s3_bucket_kubeconfig
+  acl  = "private"
+  versioning {
+    enabled = true
+  }
+
+  tags = local.tags_ks
```
deimosfr, line 10: same for tags

lib/scaleway/bootstrap/ks-workers-nodes.tf
```diff
@@ -0,0 +1,18 @@
...
+  node_type     = "DEV1-L"
+
+  # use Scaleway built-in cluster autoscaler
+  autoscaling   = true # TODO(benjaminch): use scaleway_ks_pool_autoscale variable
+  autohealing   = true
+  size          = 3 # TODO(benjaminch) : use : "{{ scw_ks_worker_node.min_size }}"
+  min_size      = 3 # TODO(benjaminch) : use : "{{ scw_ks_worker_node.min_size }}"
+  max_size      = 10 # TODO(benjaminch) : use : "{{ scw_ks_worker_node.max_size }}"
+
+  tags          =  [for i, v in local.tags_ks : "${i}=${v}"] # NOTE: Scaleway doesn't support KV style tags
```
deimosfr, line 16: same for tags

lib/scaleway/bootstrap/tf-providers.tf
```diff
@@ -0,0 +1,38 @@
...
+# }
+
+provider "kubernetes" {
+  load_config_file = "false"
+
+  host  = null_resource.kubeconfig.triggers.host
+  token = null_resource.kubeconfig.triggers.token
+  cluster_ca_certificate = base64decode(
+    null_resource.kubeconfig.triggers.cluster_ca_certificate
+  )
```
deimosfr, line 37: Do not forget vault ;)

lib/scaleway/bootstrap/ks-object-bucket.tf
benjaminch: This is ok I guess, it's already in a var and support plain dictionary.

src/object_storage/mod.rs
```diff
@@ -38,4 +39,5 @@ pub trait ObjectStorage {
 pub enum Kind {
     S3,
     Spaces,
+    ScalewayOs,
```
deimosfr, line 42: Why this name? Is this how they call it?

benjaminch: Yep, they call it `Object Storage`. If not clear enough, I can change the name to `ScalewayObjectStorage`.

tests/scaleway/scw_container_registry.rs
```diff
@@ -0,0 +1,254 @@
+extern crate test_utilities;
```
deimosfr, line 1: I'm not sure those tests are useful as they will be validated during applications tests deployments

benjaminch: Indeed, I agree, I will put an ignore once all functional tests will be up and running. But still, might be useful when functional tests will be broken to help investigate more in depth.
Let me know if ignore is ok for you ?

tests/scaleway/scw_utility_test_kubernetes_kapsule_cluster.rs
```diff
@@ -0,0 +1,109 @@
...
+use qovery_engine::cloud_provider::scaleway::kubernetes::Kapsule;
+use qovery_engine::transaction::TransactionResult;
+
+// Warning: This test shouldn't be ran by CI
+// Note: this test creates the test cluster where all application tests will be ran
+// This is not really a test but a convenient way to create the test cluster if needed to be manually created at some point.
+#[allow(dead_code)]
+#[test]
+#[ignore]
+fn create_scaleway_kubernetes_kapsule_test_cluster() {
```
deimosfr, line 17: Those tests won't work together because of generated ids. If you looks how it's done AWS side, create and delete are in the same test to avoid this issue

benjaminch: Those are not tests, but an handy function to recreate / delete test cluster on the infra, it will allow us to recreate test cluster if needed.

src/object_storage/mod.rs
deimosfr: Looks a 0 (zero) and not a O

benjaminch: I do confirm it's a O :)
Screenshot 2021-07-30 at 12 15 59

tests/scaleway/scw_container_registry.rs
benjaminch: Done, I added ignore for those kind of tests

PR #293, opened 2021-07-06
Fix: Pause action

erebe:
+ Replicas is not specified in charts so after the pause
    Helm install does not scale back statefulset to 1
  + Force replicas in chart to 1 to ensure that helm upgrade scale it back
  + Fix for MySQL were labels are not applied

https://qovery.atlassian.net/browse/DEV-927

### DIFF
--- a/lib/common/services/mongodb/templates/deployment-standalone.yaml
+++ b/lib/common/services/mongodb/templates/deployment-standalone.yaml
@@ -19,6 +19,7 @@ metadata:
 spec:
   {{- if .Values.useStatefulSet }}
   serviceName: {{ template "mongodb.serviceName" . }}
+  replicas: 1
   updateStrategy:
   {{- else }}
   strategy:
--- a/lib/common/services/redis/templates/redis-master-statefulset.yaml
+++ b/lib/common/services/redis/templates/redis-master-statefulset.yaml
@@ -19,6 +19,7 @@ spec:
       release: {{ .Release.Name }}
       role: master
   serviceName: {{ template "redis.fullname" . }}-headless
+  replicas: 1
   template:
     metadata:
       labels:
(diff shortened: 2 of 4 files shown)


lib/helm-freeze.yaml
```diff
@@ -44,10 +44,12 @@ charts:
     repo_name: loki
     version: 0.24.0
     comment: https://github.com/grafana/loki/tree/master/production/helm
+    # Chart is frozen due to custom modification see https://github.com/Qovery/engine/pull/293
```
deimosfr, line 47: please add no_sync: true to freeze them

lib/common/services/redis/templates/redis-master-statefulset.yaml
```diff
@@ -19,6 +19,7 @@ spec:
       release: {{ .Release.Name }}
       role: master
   serviceName: {{ template "redis.fullname" . }}-headless
+  replicas: 1
```
deimosfr, line 22: why don't you use a variable there?

lib/common/services/mongodb/templates/deployment-standalone.yaml
```diff
@@ -19,6 +19,7 @@ metadata:
 spec:
   {{- if .Values.useStatefulSet }}
   serviceName: {{ template "mongodb.serviceName" . }}
+  replicas: 1
```
deimosfr, line 22: why don't you use a variable there?

lib/common/services/redis/templates/redis-master-statefulset.yaml
erebe: why so ? you can't have more than one in any case. That's why they didn't bother to specify it in the first place

lib/common/services/mongodb/templates/deployment-standalone.yaml
erebe: Same than above

lib/helm-freeze.yaml
erebe: added

deimosfr (approved):

PR #298, opened 2021-07-12
Fix: Properly delete custom router

erebe:
+ Properly delete custom router when router receive a delete


src/cloud_provider/service.rs
```diff
@@ -533,6 +537,22 @@ pub fn scale_down_application(
...
+where
+    T: Router,
+{
+    send_progress_on_long_task(service, crate::cloud_provider::service::Action::Delete, || {
+        if service.has_custom_domains() {
+            let (kubernetes, environment) = match *target {
+                DeploymentTarget::ManagedServices(k, env) => (k, env),
+                DeploymentTarget::SelfHosted(k, env) => (k, env),
+            };
+            do_stateless_service_cleanup(kubernetes, environment, &service.custom_router_helm_release_name())?
```
deimosfr, line 550: can you please rename it and avoid stuffs starting with do_, it can be confusing with other methods. Thx

src/cloud_provider/aws/router.rs
```diff
@@ -488,9 +489,6 @@ impl Delete for Router {
 
     fn on_delete_error(&self, target: &DeploymentTarget) -> Result<(), EngineError> {
         warn!("AWS.router.on_delete_error() called for {}", self.name());
-
-        send_progress_on_long_task(self, crate::cloud_provider::service::Action::Delete, || {
-            delete_stateless_service(target, self, true)
-        })
+        delete_router(target, self, true)
```
deimosfr, line 492: won't this make an infinite loop?

src/cloud_provider/service.rs
erebe: done

src/cloud_provider/aws/router.rs
erebe: nop, it is a bit silly how it done ATM.
The boolean parameter is a flag to tell if it is from an error or not.
We do the same things except we fetch info from k8s (events, logs, ....) to send them in the failure.

So it is doing

Router.on_delete() -> delete_router(target, self, false)
 =>  Failure
Router.on_delete_error()  -> delete(target, self, true)
 => Failure but with infos
Run fails

deimosfr:
ready to merge @erebe ?

PR #300, opened 2021-07-16
Add do-loadbalancer-hostname to root nginx-ingress

erebe:


lib/digitalocean/bootstrap/chart_values/nginx-ingress.j2.yaml
```diff
@@ -257,6 +257,7 @@ controller:
       service.beta.kubernetes.io/do-loadbalancer-algorithm: "least_connections"
       service.beta.kubernetes.io/do-loadbalancer-size-slug: "lb-small"
       service.beta.kubernetes.io/do-loadbalancer-enable-proxy-protocol: "true"
+      service.beta.kubernetes.io/do-loadbalancer-hostname: lb.{{ managed_dns_domains_helm_format }}
```
deimosfr, line 260: is the naming valid?

erebe: that's what we have in external-dns chart

deimosfr (approved):

PR #303, opened 2021-07-29
chore: update cert manager

MacLikorne:


deimosfr:
What about lib/<providers>/bootstrap/chart_values/cert-manager.yaml? I think you missed them.

In addition if you can align on what have been done for Scaleway, would be awesome. Thx

MacLikorne:
> What about lib//bootstrap/chart_values/cert-manager.yaml? I think you missed them.
> 
> In addition if you can align on what have been done for Scaleway, would be awesome. Thx

Took a look at what is done scaleway side, the chart values file was simply removed, should I do the same ?

deimosfr:
> > What about lib//bootstrap/chart_values/cert-manager.yaml? I think you missed them.
> > In addition if you can align on what have been done for Scaleway, would be awesome. Thx
> 
> Took a look at what is done scaleway side, the chart values file was simply removed, should I do the same ?

if so, then yes, you may have to quickly adjust the chart part in the code (because the file is useless now)

MacLikorne:
> > > What about lib//bootstrap/chart_values/cert-manager.yaml? I think you missed them.
> > > In addition if you can align on what have been done for Scaleway, would be awesome. Thx
> > 
> > 
> > Took a look at what is done scaleway side, the chart values file was simply removed, should I do the same ?
> 
> if so, then yes, you may have to quickly adjust the chart part in the code (because the file is useless now)

The file was already removed with helm chart split.

PR #304, opened 2021-07-29
chore: update external dns

MacLikorne:
Works like this but in order to upgrade to 3.x.x to 5.x.x, we need to delete CRD manually and upgrade chart with new values.file


lib/aws/bootstrap/chart_values/external-dns.j2.yaml
```diff
@@ -1,19 +1,45 @@
+## @section Global parameters
```
deimosfr, line 1: please only set what's really useful in order to reduce the size and make it more visible. thx

deimosfr:
looks like you forgot DO

MacLikorne:
> looks like you forgot DO

added

PR #310, opened 2021-08-04
Enable publish service for nginx-ingress

erebe:
By enabling this option the controller is going to
use a service to know the ingress endpoint/address to use.
In our case, it is going to pick the LB hostname thus
it is going to avoid the need to add on every ingress
target annotation for external DNS with the LB hostname.
This reduce the toil to update LB endpoint when it changes,
as it is done automatically now

### DIFF
--- a/lib/aws/chart_values/nginx-ingress/nginx-ingress.j2.yaml
+++ b/lib/aws/chart_values/nginx-ingress/nginx-ingress.j2.yaml
@@ -87,7 +87,7 @@ controller:
   ## Allows customization of the external service
   ## the ingress will be bound to via DNS
   publishService:
-    enabled: false
+    enabled: true
     ## Allows overriding of the publish service to bind to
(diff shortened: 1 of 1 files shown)


benjaminch (approved):

deimosfr:
Nice catch. Questions:
1. Does it still works fine with multiple nginx ingress? (I guess yes)
2. Does it works on others cloud providers? If yes, please mange others

erebe:
> Nice catch. Questions:
> 
>     1. Does it still works fine with multiple nginx ingress? (I guess yes)

Yes, that does not change anything to the ability of running multiple ingress.It just the the address of the ingress will be the one from the svc (LB) instead of using pod ip of nginx controller

> 
>     2. Does it works on others cloud providers? If yes, please mange others
It is already ok for DO, only AWS was not set properly/consistent

PR #318, opened 2021-08-11
refactor: better DO infra support

deimosfr:


src/cloud_provider/digitalocean/application.rs
```diff
@@ -370,3 +372,83 @@ impl Listen for Application {
 pub enum StorageType {
     Standard,
 }
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum Region {
```
benjaminch, line 377: nice :)

benjaminch (approved):

src/cloud_provider/digitalocean/application.rs
deimosfr: I've learned from the best ;)

erebe (approved):
Ok, but in the long run it would be best to avoid re-using SimpleError or EngineError in code that is not part of the egnine (i.e: Do api, S3, ...) they should define their own error type to avoid colission/ handling correctly errors messages

PR #335, opened 2021-08-18
chore: cpu loki DO

pjeziorowski:


deimosfr:
It's not there anymore. I don't think you're up to date to dev branch. Also, it's related to community, not for every DO clusters and will sightly grow the défaut required resources which is not aligned with cost reduction.

PR #359, opened 2021-08-31
fix: rebuild images on dockerfile / envs change

pjeziorowski:


benjaminch (approved):

src/models.rs
```diff
@@ -317,6 +317,8 @@ impl Application {
         // https://github.com/distribution/distribution/blob/6affafd1f030087d88f88841bf66a8abe2bf4d24/reference/regexp.go#L41
         let mut hasher = DefaultHasher::new();
         self.root_path.hash(&mut hasher);
+        self.dockerfile_path.hash(&mut hasher);
+        self.environment_variables.hash(&mut hasher);
```
deimosfr, line 321: not needed for envs as helm already manages it

pjeziorowski: fixed

erebe: Are you sure env vars are not needed ? 

if env change and is used during docker build, the hash does not change and the image end-up being present in the registry. 

so we never rebuild the image even while we should have

pjeziorowski: @deimosfr ?

deimosfr: If env vars are changing, yes I'm sure it will be updated because of this: https://github.com/Qovery/engine/blob/dev/lib/aws/charts/q-application/templates/statefulset.j2.yaml#L32

If you prefer to have it there as well, no problem. @pjeziorowski can you please add a comment on those lines to explain why it's important to keep them here. Thanks

pjeziorowski: Added it back just in case + commented

PR #389, opened 2021-09-10
feat: scaleway support managed dbs

benjaminch:


lib/scaleway/services/postgresql/main.j2.tf
```diff
@@ -0,0 +1,81 @@
...
+  instance_id = scaleway_rdb_instance.postgresql_instance.id
+{%- if publicly_accessible %}
+  # By default, all IPs are authorized => 0.0.0.0/0
+  acl_rules {
+    ip = "0.0.0.0/0"
+    description = "accessible from any host"
+  }
+{%- else %}
+  # TODO(benjaminch): Allow only Scaleway's private traffic
+  acl_rules {
```
erebe, line 26: Is it possible to change the ACL afterward ?

benjaminch: this is not mentionned in their documentation, so I guess it's doable: https://developers.scaleway.com/en/products/rdb/api/#acl-rules-allowed-ips

```
ACL Rules / Allowed IPs

Access Control List Rules (ACL Rules) define which host and networks can connect to your Database Instance Endpoint. You can add, edit or delete rules from your ACLs. The initial setup of an instance allows full network access from anywhere (0.0.0.0/0).

Access control is handled directly at network-level by our load-balancers, so the Database does not need to do it itself, making the filtering more efficient, universal and relieving the Database from this task.
```

PR #395, opened 2021-09-20
fix: switch from KUBERNETES_ID to CLUSTER_ID

MacLikorne:


benjaminch (approved):

erebe:
already done when I put GRPC client in the Agent

PR #400, opened 2021-09-22
Use base64 encoded string for application env vars

erebe:
- Use base64 encoded string for env variables of applications
  - Remove External services, as they are not used by the core
    at all


evoxmusic:
FYI: "External services" are going to be re-implemented for Q4 2022.

deimosfr (approved):

PR #404, opened 2021-09-23
feat: add nat gateway and vpc peering support #Risky

deimosfr:


benjaminch (approved):

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -164,13 +179,63 @@ impl<'a> EKS<'a> {
...
+                    context.insert("eks_zone_c_subnet_blocks_public", &eks_zone_c_subnet_blocks_public);
+                }
+                VpcQoveryNetworkMode::WithoutNatGateways => {}
+            };
+            context.insert(
+                "vpc_qovery_network_mode",
+                &self.options.vpc_qovery_network_mode.clone().unwrap().to_string(),
+            );
+        } else {
+            context.insert("vpc_qovery_network_mode", &"WithoutNatGateways".to_string());
```
erebe, line 236: Use the enum.to_string(), so if you rename it one day, it will refacto automatically for you

PR #405, opened 2021-09-27
imp: SCW do not leak all image data in logs

benjaminch:


src/container_registry/scaleway_container_registry.rs
```diff
@@ -396,8 +396,8 @@ impl ContainerRegistry for ScalewayCR {
         if !force_push && self.does_image_exists(&image) {
             // check if image does exist - if yes, do not upload it again
             let info_message = format!(
-                "image {:?} found on Scaleway {} repository, container build is not required",
-                image, registry_name,
+                "image {} found on Scaleway {} repository, container build is not required",
```
deimosfr, line 399: could you add the tag and the URL may be?

benjaminch: done

erebe (approved):

PR #451, opened 2021-10-14
WIP Fix: DO stop relying on api for docker images

benjaminch:


benjaminch:
Not needed any longer.

PR #474, opened 2021-10-28
fix: log more details on cluster errors

benjaminch:
This CL also fixes tracing issue where using tracing macro spans into
closure leads to not showing logs on outputs, spans needs to be either
re-injected into the closure or make the call function handling the
logging.

This CL does the latter.


src/cmd/kubectl.rs
```diff
@@ -844,30 +844,35 @@ pub fn kubectl_exec_get_events<P>(
...
 {
     let mut environment_variables = envs;
     environment_variables.push((KUBECONFIG, kubernetes_config.as_ref().to_str().unwrap()));
 
-    let mut args = vec!["get", "event", "-A", "--sort-by='.lastTimestamp'"];
-    if !namespace.unwrap().is_empty() {
-        args = vec!["get", "event", "-n", namespace.unwrap(), "--sort-by='.lastTimestamp'"];
-    }
+    let arg_namespace = match namespace {
+        Some(n) => format!("-a {}", n),
```
deimosfr, line 855: not -a but -n

deimosfr (approved):

src/cmd/kubectl.rs
benjaminch: done

PR #475, opened 2021-10-29
lint: applying clippy fixes + warnings

benjaminch:


src/models.rs
```diff
@@ -23,7 +23,7 @@ use std::sync::Arc;
 #[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
 pub enum EnvironmentAction {
```
benjaminch, line 24: To be tested with Core, since it serialize / deserialize

src/models.rs
```diff
@@ -521,8 +518,8 @@ pub struct Route {
 
 #[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
 pub enum DatabaseMode {
```
benjaminch, line 520: tag fields to handle new casing

benjaminch:
On hold for now

PR #483, opened 2021-11-15
Feat/upgrade cluster tests

MacLikorne:


test_utilities/src/common.rs
```diff
@@ -5,16 +5,151 @@ use chrono::Utc;
...
 use std::collections::BTreeMap;
 
+pub trait Cluster<T, U> {
+    fn docker_cr_engine(context: &Context) -> Engine;
+    fn cloud_provider(context: &Context) -> Box<T>;
+    fn kubernetes_nodes() -> Vec<NodeGroups>;
+    fn kubernetes_cluster_options(secrets: FuncTestsSecrets, cluster_id: Option<String>) -> U;
+}
+
+pub trait Infrastructure {
```
benjaminch, line 34: This one rather sounds like a `Environment` trait isn't it?
If so, I would rename it to `Environment`

benjaminch: At least I am not sure to grasp diff between Environment and Infrastructure here

test_utilities/src/utilities.rs
```diff
@@ -1161,24 +1169,12 @@ pub fn test_db(
...
-            TransactionResult::Ok => assert!(true),
-            TransactionResult::Rollback(_) => assert!(false),
-            TransactionResult::UnrecoverableError(_, _) => assert!(false),
-        },
-        Kind::Scw => match scw_deploy(&context, ea, SCW_TEST_ZONE) {
-            TransactionResult::Ok => assert!(true),
-            TransactionResult::Rollback(_) => assert!(false),
-            TransactionResult::UnrecoverableError(_, _) => assert!(false),
-        },
+    match environment.deploy_environment(provider_kind.clone(), &context, &ea) {
```
benjaminch, line 1174: `environment.deploy_environment` sounds weird, there is a naming issue or fuzzy trait definition here,
if environment to be kept, then consifer rename the function to `deploy` so it's smoother => `environment.deploy(`

test_utilities/src/utilities.rs
```diff
@@ -1259,23 +1255,278 @@ pub fn test_db(
...
     }
 
-    match provider_kind.clone() {
-        Kind::Aws => match aws_delete(&context_for_delete, ea_delete) {
-            TransactionResult::Ok => assert!(true),
-            TransactionResult::Rollback(_) => assert!(false),
-            TransactionResult::UnrecoverableError(_, _) => assert!(false),
-        },
-        Kind::Do => match do_delete(&context_for_delete, ea_delete, DO_TEST_REGION) {
+    match environment_delete.delete_environment(provider_kind.clone(), &context_for_delete, &ea_delete) {
```
benjaminch, line 1258: Same about the naming

test_utilities/src/utilities.rs
MacLikorne: it could be environement deploy_charts. That's why I kept it specific

tests/aws/aws_environment.rs
```diff
@@ -496,7 +457,7 @@ fn redeploy_same_app_with_ebs() {
                 .as_str(),
         );
 
-        // Todo: make an image that check there is a mounted disk
+        let storage_size: u8 = 10;
```
benjaminch, line 460: Maybe use u16 directly here, avoid the cast below.

benjaminch (approved):

PR #485, opened 2021-11-16
feat: adding SCW env pause + fixed test issue

deimosfr:
Tests were not working as expected, there was a bug in the function naming as well, and added SCW support


MacLikorne (approved):

src/cmd/kubectl.rs
```diff
@@ -1033,6 +1046,59 @@ where
...
+{
+    let condition_format = format!(
+        "for={}",
+        match condition {
+            PodCondition::Delete => format!("{:?}", &condition).to_lowercase(),
+            _ => format!("condition={:?}", &condition).to_lowercase(),
+        }
+    );
+
+    let mut _envs = Vec::with_capacity(envs.len() + 1);
```
benjaminch, line 1077: shouldn't it be `_envs` since this one is being used?

benjaminch (approved):

PR #500, opened 2021-11-19
feat: use subdomains for dns and tls

benjaminch:
Ticket DEV-996


src/models.rs
```diff
@@ -1131,3 +1134,167 @@ impl Metadata {
...
+pub trait ToHelmString {
+    fn to_helm_format_string(&self) -> String;
+}
+
+/// Represents a domain, just plain domain, no protocol.
+/// eq. `test.com`, `sub.test.com`
+#[derive(Clone)]
+pub struct Domain {
+    raw: String,
+    top_domain: String,
```
erebe, line 1151: OPT: call it root domain to be consistent ?

erebe (approved):

src/models.rs
benjaminch: Yep, you are right, thanks :)

test_utilities/src/common.rs
```diff
@@ -660,3 +667,64 @@ pub fn environment_only_http_server_router(context: &Context, organization_id: &
...
     }
 }
+
+pub fn deploy_upgrade_destroy_infra_and_env<'a>(
+    tx: &'a mut Transaction<'a>,
+    kubernetes: &'a dyn Kubernetes,
+    environment_action: &'a EnvironmentAction,
+) {
+    // Deploy infrastructure
+    if let Err(err) = tx.create_kubernetes(kubernetes) {
```
MacLikorne, line 677: With the transaction commit behavior, this func will do a lot of useless stuff as we talked yesterday

src/models.rs
```diff
@@ -1131,3 +1134,170 @@ impl Metadata {
...
+
+impl Domain {
+    pub fn new(raw: String) -> Self {
+        // TODO(benjaminch): This is very basic solution which doesn't take into account
+        // some edge cases such as: "test.co.uk" domains
+        let sep: &str = ".";
+        let items: Vec<String> = raw.split(sep).map(|e| e.to_string()).collect();
+        let items_count = raw.matches(sep).count() + 1;
+        let top_domain: String = match items_count > 2 {
+            true => items.iter().skip(items_count - 2).join("."),
```
deimosfr, line 1162: wasn't it more simple to do: format!("{}.{}", items[-2], items[-1])?

test_utilities/src/cloudflare.rs
```diff
@@ -1,15 +1,24 @@
 use crate::utilities::FuncTestsSecrets;
 use qovery_engine::dns_provider::cloudflare::Cloudflare;
-use qovery_engine::models::Context;
+use qovery_engine::models::{Context, Domain};
 
-pub fn dns_provider_cloudflare(context: &Context) -> Cloudflare {
+pub enum CloudflareDomain {
```
deimosfr, line 5: Why Default is useful? I don't see any reasons where it's used instead for dropping an error. Why not using a string instead in dns_provider_cloudflare?

deimosfr:
Good luck for the rebase, you can thanks @MacLikorne

test_utilities/src/common.rs
benjaminch: Done :)

test_utilities/src/cloudflare.rs
benjaminch: I renamed it to be `ClusterDomain` instead.

- Default allows to say to the engine to let Cloudflare and DNS providers to use their default domain given via secrets or whatever, whereas custom indicate to use the custom one, agin, a bit more clear when you see it in code later on.

- We should try to avoid setting everything inside a string as much as we can and leverage types :)

- As of why it's better using an enum rather than a &str is that types carry information and it's easier to maniupulate and know what it should represent.

src/models.rs
benjaminch: There is for sure a better way / optimized from what I did,. but in your example, you are fetching array two times (it shouldn't be a big array, so not an issue).
Also, there is in my knowledge no such Python like negative indexing looping thing in Rust. But maybe I am wrong.

benjaminch:
> Good luck for the rebase, you can thanks @MacLikorne

Enjoying it a lot :)

PR #503, opened 2021-11-23
feat: introduce new error handling models

benjaminch:
Ticket: DEV-1004

### DIFF
--- a/src/errors/mod.rs
+++ b/src/errors/mod.rs
@@ -1,0 +1,234 @@
+pub mod io;
+
+extern crate url;
+
+use crate::cloud_provider::Kind;
+use url::Url;
+
+pub struct SimpleError {
+    message: String,
+    message_safe: String,
+}
+
+#[derive(Debug)]
+pub enum LogLevel {
+    Debug,
+    Info,
+    Warning,
+    Error,
+    Critical,
+}
+
+#[derive(Debug)]
+pub enum Stage {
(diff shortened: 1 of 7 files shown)


src/errors/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    message_safe: String,
+}
+
+#[derive(Debug)]
+pub enum LogLevel {
+    Debug,
+    Info,
+    Warning,
+    Error,
+    Critical,
```
erebe, line 19: What the difference between Error and Critical ?

src/errors/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    Upgrade,
+    Delete,
+}
+
+#[derive(Debug)]
+pub enum EnvironmentStep {
+    Build,
+    Deploy,
+    Update,
+    Delete,
```
erebe, line 41: Pause ?

MacLikorne (approved):

src/errors/mod.rs
benjaminch: I would say Critical will fail the process instantly, but maybe Error should too, so I guess I can remove Critical.

src/errors/mod.rs
benjaminch: Sure ! Thanks, I wasn't exhaustive, this list will grow along migration :)

PR #507, opened 2021-11-25
fix: grow default AWS timeout to reduce random AWS issues

deimosfr:


lib/aws/bootstrap/eks-master-cluster.j2.tf
```diff
@@ -53,6 +53,12 @@ resource "aws_eks_cluster" "eks_cluster" {
 
   tags = local.tags_eks
 
+  # grow this value to reduce random AWS API timeouts
+  timeouts {
+    create = "60 minutes"
```
MacLikorne, line 58: Is the create for the step (cluster + workers ) ? If yes it still could be short.

deimosfr: Nope, it's just the cluster creation. Workers is another kind of resource

benjaminch (approved):

PR #511, opened 2021-11-30
Add new field ports for applications

erebe:


src/models.rs
```diff
@@ -160,6 +160,22 @@ fn default_root_path_value() -> String {
...
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
+pub enum Protocol {
+    HTTP,
+}
+
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
+pub struct Port {
+    pub id: String,
+    pub long_id: uuid::Uuid,
+    pub port: u16,
```
benjaminch, line 172: maybe try to find another name, because it feels weird to have Port.port, maybe `number`?

src/models.rs
```diff
@@ -160,6 +160,22 @@ fn default_root_path_value() -> String {
     "/".to_string()
 }
 
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
+pub enum Protocol {
+    HTTP,
+}
+
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
+pub struct Port {
```
benjaminch, line 169: Maybe create a constructor on `Port`, it will allow you to omit `publicly_accessible` field since it can be computed from `public_port.is_some()`.

src/models.rs
erebe: done

src/models.rs
erebe: This is the struct/payload received by NATS/core

benjaminch (approved):

PR #522, opened 2021-12-10
feat: adding more logger capabilities

benjaminch:


src/events/io.rs
```diff
@@ -1 +1,187 @@
...
+            events::Stage::Environment(step) => Stage::Environment(EnvironmentStep::from(step)),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum InfrastructureStep {
+    Instantiate,
+    Create,
```
deimosfr, line 73: What is the difference between Instantiate and create? Isn't it deploy missing?

benjaminch: Instanciate is when we purely pop new K8s object (constructor). This  can be called for any step.

In my understanding there is no such thing as deploy, everything is a deploy for infra, it's either create, upgrade or delete.

I might miss some statuses such as Resume or stuff like that I will add those along when calling this new logger.

PR #524, opened 2021-12-13
Refacto git operations

erebe:
- Properly set login/password to repository url
    They were not correctly escaped before

  - Use git reset --hard to checkout because we were trying
    to change (badly) the HEAD which is what reset does

  - Simplify + Handle all errors instead of swallowing them

  - Add some tests

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,91 +1,147 @@
 use std::path::Path;
 
-use git2::build::RepoBuilder;
-use git2::{Error, Oid, Repository};
+use git2::build::{CheckoutBuilder, RepoBuilder};
+use git2::ResetType::Hard;
+use git2::{Error, Repository};
+use url::Url;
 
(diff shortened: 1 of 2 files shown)


benjaminch (approved):

src/git.rs
```diff
@@ -1,91 +1,145 @@
...
-            repository_url.replace("https://", "")
-        ),
-        None => repository_url.to_string(),
-    };
+    let mut url = Url::parse(repository_url)
+        .map_err(|err| Error::from_str(format!("Invalid repository url {}: {}", repository_url, err).as_str()))?;
 
-    RepoBuilder::new().clone(final_repository_url.as_str(), into_dir.as_ref())
-}
+    if url.scheme() != "https" {
```
deimosfr, line 20: what if git?

erebe: not supported,  the core never send a git@ url

deimosfr (approved):

PR #525, opened 2021-12-14
Add autoscaler for applications

erebe:
- Transition to min/max instances instead of total instances


src/cloud_provider/aws/databases/mongodb.rs
```diff
@@ -141,7 +141,11 @@ impl Service for MongoDB {
         self.total_ram_in_mib
     }
 
-    fn total_instances(&self) -> u16 {
+    fn min_instances(&self) -> u32 {
+        1
```
benjaminch, line 145: Maybe stick a comment here explaining why it's 1 as min and max.
If those methods are not relevant to DB, maybe move those out into a new trait which won't be implemented by DB ? Something like `ScalableService` ?

MacLikorne (dismissed):

src/cloud_provider/aws/databases/mongodb.rs
erebe: Imho, we should remove in the long run the Service trait.
It makes a db and an app share the same shape, and the more we will customize each of them, the more it will stretch from each others

benjaminch: Perfect :)

PR #527, opened 2021-12-14
fix: kubernetes version upgrade

MacLikorne:


src/cmd/kubectl.rs
```diff
@@ -1049,3 +1057,14 @@ where
 
     Ok(result)
 }
+
+pub fn get_all_pdbs<P>(kubernetes_config: P, envs: Vec<(&str, &str)>) -> Result<PDB, SimpleError>
```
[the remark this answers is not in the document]
MacLikorne: done

benjaminch (dismissed):

src/cmd/structs.rs
```diff
@@ -408,6 +408,33 @@ pub struct SVCSpec {
     pub svc_type: String,
 }
 
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PDB {
```
benjaminch, line 413: Can you please add a deserialize test please ? We've seen some issues lately when bumping kubectl version, it would allow us preventing that from happening in the future.

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -1345,6 +1345,24 @@ impl<'a> Kubernetes for EKS<'a> {
             }
         }
 
+        match self.delete_crashlooping_pods(
```
benjaminch, line 1348: Replace with a if let Err()

src/cloud_provider/kubernetes.rs
```diff
@@ -197,6 +202,63 @@ pub trait Kubernetes: Listen {
 
         return Ok(path);
     }
+
+    fn delete_crashlooping_pods(
```
benjaminch, line 206: Tl me this one should be part of kubectl comd file.

benjaminch: Also, always good IMO to return deleted pods in results.

PR #529, opened 2021-12-15
Refactor: k8s get pods method to allow optional namespace nor selector

benjaminch:


src/cmd/kubectl.rs
```diff
@@ -715,20 +715,31 @@ where
...
 ) -> Result<KubernetesList<KubernetesPod>, SimpleError>
 where
     P: AsRef<Path>,
 {
-    kubectl_exec::<P, KubernetesList<KubernetesPod>>(
-        vec!["get", "pod", "-o", "json", "-n", namespace, "-l", selector],
-        kubernetes_config,
-        envs,
-    )
+    let mut cmd_args = vec!["get", "pod", "-o", "json"];
```
MacLikorne, line 727: To have a list i think you need to set podS instead of pod

benjaminch: Indeed :)

PR #530, opened 2021-12-15
feat: allowing to delete k8s crashlooping pods

benjaminch:


src/cmd/kubectl.rs
```diff
@@ -1017,6 +1017,88 @@ where
...
+where
+    P: AsRef<Path>,
+{
+    let restarted_min = restarted_min_count.unwrap_or_else(|| 10usize);
+    let pods = kubectl_exec_get_pods(kubernetes_config, namespace, selector, envs)?;
+
+    Ok(pods
+        .items
+        .into_iter()
+        .filter(|e| {
```
MacLikorne, line 1045: does e stand for pod ? If you yes, I'd suggest to switch to an explicit var

src/cmd/kubectl.rs
```diff
@@ -1017,6 +1017,88 @@ where
...
+                    ));
+                }
+
+                pods.items[0].clone()
+            }
+            Err(e) => return Err(e),
+        };
+
+    kubectl_exec(
+        vec!["delete", "pod", pod_to_be_deleted.metadata.name.as_str()],
```
MacLikorne, line 1094: If i remember well you have specify namespace and add an arg for pod name:

delete pod -n <namespace> --name <pod_name>

benjaminch: Indeed, thanks, --name is not an option, but namespace is required.

src/cmd/kubectl.rs
benjaminch: Yep, ok !

deimosfr (approved):

PR #531, opened 2021-12-15
feat: support aws zones

deimosfr:


Cargo.toml
```diff
@@ -28,6 +28,7 @@ uuid = { version = "0.8", features = ["v4", "serde"] }
 url = "2.2.2"
 function_name = "0.2.0"
 thiserror = "1.0.30"
+regex = "1.5.4"
```
benjaminch, line 31: Aie aie aie

src/cloud_provider/aws/regions.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    pub fn to_string(&self) -> String {
+        let enum_name = format!("{}", self);
+        format!("{}", enum_name)
+    }
+
+    pub fn to_aws_format(&self) -> String {
+        let enum_name = format!("{}", self);
+        let re = Regex::new(r"^(.+)([A-Z])(.+)(\d)").expect("error while building the regex for AWS zones");
```
benjaminch, line 138: :/ can we avoid regex here?

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -229,6 +232,25 @@ impl<'a> EKS<'a> {
 
         let format_ips =
             |ips: &Vec<String>| -> Vec<String> { ips.iter().map(|ip| format!("\"{}\"", ip)).collect::<Vec<_>>() };
+        let format_zones = |zones: &Vec<AwsZones>| -> Vec<String> {
+            zones
+                .iter()
+                .map(|zone| format!("\"{}\"", zone.to_string()))
```
benjaminch, line 238: Maybe you can implement newly introduced traits, it will bring more clarity and can be reused from other places `pub trait ToTerraformString` or / and `pub trait ToHelmString ` https://github.com/Qovery/engine/blob/739e3d12e9aa7bbcbe58be754d7344d066663cc7/src/models.rs#L1216

src/cloud_provider/aws/regions.rs
```diff
@@ -0,0 +1,301 @@
...
+        let sanitized_name = re.replace(enum_name.as_str(), |caps: &Captures| {
+            format!("{}-{}{}-{}", &caps[1], &caps[2], &caps[3], &caps[4]).to_lowercase()
+        });
+        format!("{}", sanitized_name)
+    }
+
+    pub fn aws_region_from_str(s: &str) -> Option<AwsRegion> {
```
benjaminch, line 145: Why not implementing `From<&str>` trait here?

src/cloud_provider/aws/regions.rs
deimosfr: Not as  I know

Cargo.toml
deimosfr: <3

deimosfr: What's the problem?

benjaminch: Not a big fan of regex, comes with an entire ecosystem, if we have only a very small use case and can find another alternative, that's my way to go, but if there is no other solution, then ... but IMO, we should try to avoid it while we can.

src/cloud_provider/aws/regions.rs
benjaminch: ok

benjaminch (dismissed):

MacLikorne (dismissed):

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -243,6 +263,25 @@ impl<'a> EKS<'a> {
 
         let format_ips =
             |ips: &Vec<String>| -> Vec<String> { ips.iter().map(|ip| format!("\"{}\"", ip)).collect::<Vec<_>>() };
+        let format_zones = |zones: &Vec<AwsZones>| -> Vec<String> {
```
erebe, line 266: Why a clojure/lambda ? You are invoking it only once

src/cloud_provider/aws/mod.rs
```diff
@@ -121,6 +125,10 @@ impl CloudProvider for AWS {
         }
     }
 
+    fn zones(&self) -> Vec<String> {
```
erebe, line 128: Return a reference, you are always cloning and after modifying the zone, which re-duplicate memory

src/cloud_provider/aws/regions.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    pub fn to_string(&self) -> String {
+        let enum_name = format!("{}", self);
+        format!("{}", enum_name)
+    }
+
+    pub fn to_aws_format(&self) -> String {
+        let enum_name = format!("{}", self);
+        let re = Regex::new(r"^(.+)([A-Z])(.+)(\d)").expect("error while building the regex for AWS zones");
```
erebe, line 181: Remove the regex  and create a custom mapper with a match on the enum.
Nobody will expect to break this function if it renames an enum.
also it will remove the need for the regex

src/cloud_provider/aws/regions.rs
```diff
@@ -0,0 +1,373 @@
...
+    RegionNotFound,
+    RegionNotSupported,
+    ZoneNotFound,
+    ZoneNotSupported,
+}
+
+impl AwsZones {
+    pub fn to_string(&self) -> String {
+        let enum_name = format!("{}", self);
+        let re = Regex::new(r"^(.+)([A-Z])(.+)(\d)(.+)").expect("error while building the regex for AWS zones");
```
erebe, line 278: same

test_utilities/src/common.rs
```diff
@@ -1194,7 +1198,7 @@ pub fn get_environment_test_kubernetes<'a>(
                     uuid::Uuid::new_v4(),
                     format!("qovery-{}", context.cluster_id()),
                     DO_KUBERNETES_VERSION.to_string(),
-                    Region::from_str(
+                    DoRegion::from_str(
                         secrets
                             .clone()
```
benjaminch, line 1203: nit: can be `as_ref()`

src/cloud_provider/digitalocean/network/vpc.rs
```diff
@@ -248,11 +248,11 @@ mod tests_do_vpcs {
     #[test]
     fn check_reserved_subnets() {
         // if not reserved
-        assert_eq!(is_do_reserved_vpc_subnets(Region::Frankfurt, "192.168.0.0/24"), false);
+        assert_eq!(is_do_reserved_vpc_subnets(DoRegion::Frankfurt, "192.168.0.0/24"), false);
```
benjaminch, line 251: Can be `assert_eq!(!is_do_reserved_vpc_subnets(DoRegion::Frankfurt, "192.168.0.0/24"));`

src/cloud_provider/aws/kubernetes/mod.rs
deimosfr: To make it more clear to read

benjaminch (dismissed):

src/cloud_provider/digitalocean/mod.rs
```diff
@@ -115,6 +115,10 @@ impl CloudProvider for DO {
         }
     }
 
+    fn zones(&self) -> &Vec<String> {
+        todo!()
```
erebe, line 119: Remove the todo!()

MacLikorne (approved):

src/cloud_provider/digitalocean/mod.rs
deimosfr: What do you suggest to replace with?

PR #533, opened 2021-12-16
feat: changing way to detect crashlooping pods

benjaminch:


MacLikorne (approved):

deimosfr (dismissed):

src/cmd/kubectl.rs
```diff
@@ -1036,14 +1037,14 @@ pub fn kubectl_get_crash_looping_pods<P>(
...
+    let restarted_min = restarted_min_count.unwrap_or(10usize);
     let pods = kubectl_exec_get_pods(kubernetes_config, namespace, selector, envs)?;
 
     Ok(pods
         .items
         .into_iter()
         .filter(|pod| {
-            pod.status.phase != KubernetesPodStatusPhase::Running
+            pod.status.reason == KubernetesPodStatusReason::BackOff
                 && pod.status.container_statuses.as_ref().is_some()
```
deimosfr, line 1048: is it useful to check the status then? I'm not sure, if it's backofff, then...

benjaminch: It is in case we want to kill if N restarts

deimosfr: May be you can add a comment, because it's not obvious

benjaminch: Added a comment, hpefuly it's clearer, let me know.

PR #535, opened 2021-12-20
feat: adding rds disk encryption option

deimosfr:


deimosfr:
Waiting for core team validation before merging

src/models.rs
```diff
@@ -609,6 +609,7 @@ pub struct Database {
     pub disk_size_in_gib: u32,
     pub database_instance_type: String,
     pub database_disk_type: String,
+    pub database_disk_encryption: bool,
```
erebe, line 612: Why prefix the field with database_ ? We are already in the database object

deimosfr: To have consistency with others. But I agree we can remove it as it's already in databse

benjaminch (approved):

erebe:
Please ensure we set in the core DB all flag encryption to false when this PR is merged

PR #541, opened 2021-12-31
fix: do connection loss

MacLikorne:


src/cloud_provider/digitalocean/kubernetes/mod.rs
```diff
@@ -1100,6 +1101,25 @@ impl<'a> DOKS<'a> {
...
+        if let Err(e) = self.spaces.empty_bucket(self.kubeconfig_bucket_name().as_str()) {
+            return Err(EngineError::new(
+                EngineErrorCause::Internal,
+                self.engine_error_scope(),
+                self.context().execution_id(),
+                e.message,
+            ));
+        }
+
+        if let Err(e) = self.spaces.empty_bucket(self.logs_bucket_name().as_str()) {
```
benjaminch, line 1114: Should we trash logs ? Don't we want to keep those ?

MacLikorne: It's on cluster deletion. In Scaleway, it's deleted, that's why I reproduce it there but without deleting buckets (DO issue, you know). But yeah keeping logs could be an idea.

benjaminch: Ok ok ! Let's be consistent then ;)

benjaminch (approved):

src/cloud_provider/digitalocean/kubernetes/mod.rs
MacLikorne: Done !

src/cloud_provider/digitalocean/kubernetes/mod.rs
```diff
@@ -1100,6 +1099,16 @@ impl<'a> DOKS<'a> {
             }
         }
 
+        info!("Empty Qovery managed object storage buckets");
+        if let Err(e) = self.spaces.empty_bucket(self.kubeconfig_bucket_name().as_str()) {
```
benjaminch, line 1103: I guess you can also use the `delete` bucket function, bqased on spaces stratgey, which should be `empty`, deleting the bucket will lead to only empty it => https://github.com/Qovery/engine/blob/729e96c69ea6a8a6f1426bb0bc201e5bed338b83/src/object_storage/spaces.rs#L275

benjaminch (dismissed):

src/cloud_provider/digitalocean/kubernetes/mod.rs
MacLikorne: Since it's DO bucket with the deletion issue and local tests always use the same name, it's mandatory to only empty it. Pleco will delete empty buckets anyway is not used for a long time.

benjaminch (approved):

PR #544, opened 2022-01-04
Add support for ssh key during git submodule

erebe:
You can ignore git guardian

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,36 +1,45 @@
 use std::path::Path;
 
 use git2::build::{CheckoutBuilder, RepoBuilder};
+use git2::ErrorCode::Auth;
 use git2::ResetType::Hard;
-use git2::{Error, Repository};
+use git2::{Cred, CredentialType, Error, RemoteCallbacks, Repository, SubmoduleUpdateOptions};
 use url::Url;
 
-pub struct Credentials {
-    pub login: String,
-    pub password: String,
-}
+// Credentials callback is called endlessly until the server return Auth Ok (or a definitive error)
+// If auth is denied, it up to us to return a new credential to try different auth method
+// or an error to specify that we have exhausted everything we are able to provide
+fn authentication_callback<'a>(
+    get_credentials: &'a impl Fn() -> Vec<(CredentialType, Cred)>,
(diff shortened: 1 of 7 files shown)


src/git.rs
```diff
@@ -50,18 +59,77 @@ pub fn checkout(repo: &Repository, commit_id: &str) -> Result<(), Error> {
...
+    get_credentials: &impl Fn() -> Vec<(CredentialType, Cred)>,
+) -> Result<Repository, Error>
+where
+    P: AsRef<Path>,
+{
+    let url = Url::parse(repository_url)
+        .map_err(|err| Error::from_str(format!("Invalid repository url {}: {}", repository_url, err).as_str()))?;
+
+    if url.scheme() != "https" {
+        return Err(Error::from_str("Repository URL have to start with https://"));
```
MacLikorne, line 74: Isn't it a bit weird to only allow ssh for submodule but not main repo ?

erebe: (This line verify that it is https and not just http, as we send credentials along)

For your general question,
For submodule, you can clone them via https:// (modulo it is public or that your credentials works on it) or via ssh (but only with ssh key, as it is the only auth method possible with this scheme)

For cloning the main repo, we allow only `https://` just to avoid mistake, credentials passed in the git_reposiory object should be valid and as login/pass with auth token is only possible with httpx scheme, I force it

MacLikorne: Ok got it ! Thx

MacLikorne (approved):

src/git.rs
```diff
@@ -50,18 +59,77 @@ pub fn checkout(repo: &Repository, commit_id: &str) -> Result<(), Error> {
...
+    into_dir: P,
+    get_credentials: &impl Fn() -> Vec<(CredentialType, Cred)>,
+) -> Result<Repository, Error>
+where
+    P: AsRef<Path>,
+{
+    let url = Url::parse(repository_url)
+        .map_err(|err| Error::from_str(format!("Invalid repository url {}: {}", repository_url, err).as_str()))?;
+
+    if url.scheme() != "https" {
```
benjaminch, line 73: what a shame there is no `is_secure()` :(

benjaminch (approved):

PR #546, opened 2022-01-04
feat: allowing to activate sticky sessions on router

benjaminch:
Ticket: ENG-1086


erebe (approved):

deimosfr (dismissed):
Next time, please split the real purpose vs refacto without any relation

erebe (approved):

PR #548, opened 2022-01-05
fix local docker build cache while not using env vars in the Dockerfile

evoxmusic:
Hi, here is a fix to better use the Docker cache while building an app with a Dockerfile. 

## Motivation
My main motivation is to speed up app deployment with Qovery. Some users were complaining about the very long time to deploy an app. 

## What does this fix do
My fix looks at the environment variables that are used in the Dockerfile to inject only those that are used. Then Docker will not unnecessarily invalidate the cache each time we deploy an app. I do expect to have a cache hit of at least 80% with this fix.


src/build_platform/local_docker.rs
```diff
@@ -58,6 +58,19 @@ impl LocalDocker {
...
     }
 
+    /// Read Dockerfile content from location path and return an array of bytes
+    fn get_dockerfile_content(&self, dockerfile_path: &str) -> Result<Vec<u8>, EngineError> {
+        match fs::read(dockerfile_path) {
+            Ok(bytes) => Ok(bytes),
+            Err(err) => {
+                let error_msg = format!("Can't read Dockerfile '{}'", dockerfile_path);
+                error!("{:?}", error_msg);
+                error!("{:?}", err);
```
benjaminch, line 68: Better to combine those two making sure log is not far, very handy when a lot of stuff is running concurently:
```
error!("Can't read Dockerfile '{}', error: {:?}", dockerfile_path, err);
```

src/build_platform/local_docker.rs
```diff
@@ -84,13 +97,25 @@ impl LocalDocker {
 
         docker_args.extend(vec!["-f", dockerfile_complete_path, "-t", name_with_tag.as_str()]);
 
+        let dockerfile_content = self.get_dockerfile_content(dockerfile_complete_path)?;
+        let env_var_args = match docker::match_used_env_var_args(env_var_args, dockerfile_content) {
+            Ok(env_var_args) => env_var_args,
+            Err(err) => {
+                let error_msg = format!("Can't extract env vars from Dockerfile '{}'", dockerfile_complete_path);
```
benjaminch, line 104: Same, combine logs.

src/docker.rs
```diff
@@ -0,0 +1,139 @@
+use std::collections::HashSet;
```
benjaminch, line 1: Silly question, but since this file seems to be used only in `src/build_platform/local_docker.rs`, would it makes sense to put it inside the same directory `src/build_platform/`?

src/docker.rs
```diff
@@ -0,0 +1,139 @@
...
+/// ```
+///
+/// will return a vector of "foo" and "bar" stings
+pub fn extract_dockerfile_args(dockerfile_content: Vec<u8>) -> Result<HashSet<String>, Utf8Error> {
+    let lines = std::str::from_utf8(dockerfile_content.as_slice())?;
+    let lines = lines.lines();
+
+    let used_args = lines
+        .into_iter()
+        .filter(|line| line.to_uppercase().trim().starts_with("ARG"))
```
erebe, line 22: Replace "ARG" by "ARG " (with a space at the end) to avoid matching "ARGENT" "ARGUMENT" ...
I would remove the trim also, as it is a command it cannot be indented out, so it would avoid matching unecessary stuff

src/docker.rs
evoxmusic: I was wondering the same. It makes sense for me to put it in the `build_platform` module and move it down if it could be used elsewhere in the future 👍

src/docker.rs
evoxmusic: It makes sense. Thank you

src/build_platform/local_docker.rs
evoxmusic: thank you

erebe (approved):

PR #551, opened 2022-01-06
feat(ENG_1044): node check impl

MacLikorne:


src/cloud_provider/kubernetes.rs
```diff
@@ -704,6 +732,38 @@ pub fn check_kubernetes_has_enough_resources_to_deploy_environment(
...
+) -> Result<(), SimpleError>
+where
+    P: AsRef<Path>,
+{
+    return match kubectl_exec_get_node(kubernetes_config, envs) {
+        Err(e) => {
+            error!("Can't get nodes before services deployments");
+            Err(e)
+        }
+        Ok(nodes) => match nodes.items.len() < max_nodes * 80 / 100 {
```
deimosfr, line 749: please add () to avoid any bad missunderstanding

src/cloud_provider/kubernetes.rs
```diff
@@ -704,6 +732,38 @@ pub fn check_kubernetes_has_enough_resources_to_deploy_environment(
...
+) -> Result<(), SimpleError>
+where
+    P: AsRef<Path>,
+{
+    return match kubectl_exec_get_node(kubernetes_config, envs) {
+        Err(e) => {
+            error!("Can't get nodes before services deployments");
+            Err(e)
+        }
+        Ok(nodes) => match nodes.items.len() < (max_nodes * 80 / 100) {
```
erebe, line 749: doesn't it break autoscaler ?
if you have a cluster with [3 min, 10 max] nodes,
and say you are at 3 nodes ATM, you will never be able to scale your cluster no ?

erebe, line 749: if you are already at max node, let's say 3;3. you will never be able to deploy something no ?

MacLikorne: yes that's the point, if more than 80% of nodes capacity is used, it's stuck

erebe: ok for infra, but for env you will never be able to deploy something has you will have 3 nodes, but they will be empty

deimosfr: If you're 3/3 yes, but you can be sure that at least 50% of the time you won't be able to deploy. So, the user will just have to upgrade the max capacity. That won't mean the cluster capacity will grow, but we reduce issues due to this point.

PR #556, opened 2022-01-11
feat(ENG_1085): test db access switch

MacLikorne:


test_utilities/src/common.rs
```diff
@@ -46,6 +46,11 @@ pub enum ClusterDomain {
     Custom(String),
 }
 
+pub enum DbTestType {
+    Classic,
+    WithAccessSwitch,
```
deimosfr, line 51: WithAccessSwitch is not clear about what it does, I suggest renaming it

PR #558, opened 2022-01-12
fix build args injection while building docker image

evoxmusic:
Hi, here is a fix for build args injection while building a Docker image. It didn't work because of the bad assumptions of the format of `env_var_args`.

### DIFF
--- a/src/build_platform/docker.rs
+++ b/src/build_platform/docker.rs
@@ -43,14 +43,16 @@ pub fn match_used_env_var_args(
     let used_args = extract_dockerfile_args(dockerfile_content)?;
 
     // match env var args and dockerfile env vargs
-    let env_var_arg_keys = env_var_args.iter()
+    let env_var_arg_keys = env_var_args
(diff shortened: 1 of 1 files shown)


benjaminch (approved):

src/build_platform/docker.rs
```diff
@@ -42,7 +43,15 @@ pub fn match_used_env_var_args(
     let used_args = extract_dockerfile_args(dockerfile_content)?;
 
     // match env var args and dockerfile env vargs
-    Ok(HashSet::from_iter(env_var_args)
+    let env_var_arg_keys = env_var_args
+        .iter()
+        .map(|env_var| {
+            let x = env_var.split("=").collect::<Vec<&str>>();
+            x.get(0).unwrap_or(&"").to_string()
```
benjaminch, line 50: maybe `next()` will be easier / clearer here.

benjaminch (approved):

PR #561, opened 2022-01-17
feat: upgrade metrics server

MacLikorne:


src/cloud_provider/digitalocean/kubernetes/helm_charts.rs
```diff
@@ -428,15 +428,8 @@ pub fn do_helm_charts(
         chart_info: ChartInfo {
             name: "metrics-server".to_string(),
             path: chart_path("common/charts/metrics-server"),
+            values_files: vec![chart_path("chart_values/metrics-server.yaml")],
             values: vec![
-                ChartSetValue {
```
deimosfr, line 432: why did you remove it?

src/cmd/kubectl.rs
```diff
@@ -1239,3 +1239,14 @@ where
         envs,
     )
 }
+
+pub fn kubernetes_get_all_hpas<P>(kubernetes_config: P, envs: Vec<(&str, &str)>) -> Result<HPA, SimpleError>
```
deimosfr, line 1243: Please take a look on how other similar cmd works to select namespaces. It would be simpler for later usage in a single namespace. Thx

test_utilities/src/common.rs
```diff
@@ -1510,3 +1544,34 @@ pub fn cluster_test(
...
+where
+    P: AsRef<Path>,
+{
+    let result = kubernetes_get_all_hpas(kubernetes_config, envs);
+
+    match result {
+        Ok(hpas) => {
+            for hpa in hpas.items.expect("No hpa item").into_iter() {
+                if hpa.metadata.annotations.is_some() {
+                    if hpa
```
deimosfr, line 1558: can't you check something working? because if it's another error, we may have issues

src/cloud_provider/digitalocean/kubernetes/helm_charts.rs
MacLikorne: Because it was inconsistent. Replaced for all by a values file. If you prefer I can put it all in rust instead of yaml.

deimosfr (approved):

PR #569, opened 2022-01-20
feat: aws s3 move to rusoto

benjaminch:
+ adding functional tests


src/object_storage/s3.rs
```diff
@@ -1,39 +1,148 @@
...
+                bucket_name,
+                message.unwrap_or_else(|| "unknown error".to_string())
+            );
+            return Err(self.engine_error(EngineErrorCause::Internal, message));
+        }
+
+        let s3_client = self.get_s3_client();
+
+        // make sure to delete all bucket content before trying to delete the bucket
+        let objects_to_be_deleted = match block_on(s3_client.list_objects(ListObjectsRequest {
```
deimosfr, line 106: We've seen with Loki how painful it is to wait for a long time for some s3 buckets. May be we have another solution to speed this up?

src/object_storage/s3.rs
```diff
@@ -60,48 +169,138 @@ impl ObjectStorage for S3 {
...
+        if let Err(e) = block_on(s3_client.put_bucket_tagging(PutBucketTaggingRequest {
+            bucket: bucket_name.to_string(),
+            expected_bucket_owner: None,
+            tagging: Tagging {
+                tag_set: vec![
+                    Tag {
+                        key: "CreationDate".to_string(),
+                        value: format!("{}", creation_date.to_rfc3339()),
+                    },
+                    Tag {
```
deimosfr, line 215: I prefer not to see this. Please prepare the request before, and add the TTL if necessary. If we run Pleco at some points, 0 means immediatly, which is dangerous

test_utilities/src/aws.rs
```diff
@@ -20,14 +20,15 @@ use crate::cloudflare::dns_provider_cloudflare;
...
-pub const AWS_TEST_REGION: &str = "eu-west-3";
+pub const AWS_REGION_FOR_S3: AwsRegion = AwsRegion::EuWest3;
+pub const AWS_TEST_REGION: AwsRegion = AwsRegion::EuWest3;
 pub const AWS_KUBERNETES_MAJOR_VERSION: u8 = 1;
 pub const AWS_KUBERNETES_MINOR_VERSION: u8 = 20;
 pub const AWS_KUBERNETES_VERSION: &'static str =
     formatcp!("{}.{}", AWS_KUBERNETES_MAJOR_VERSION, AWS_KUBERNETES_MINOR_VERSION);
 pub const AWS_DATABASE_INSTANCE_TYPE: &str = "db.t3.micro";
 pub const AWS_DATABASE_DISK_TYPE: &str = "gp2";
+pub const AWS_RESOURCE_TTL_IN_SECONDS: u32 = 7200;
```
deimosfr, line 31: Please use already existing Vault value for TTL and not another one

deimosfr (changes requested):

src/object_storage/s3.rs
benjaminch: Unfortunately there is no such things via Rusoto, no option to delete a non empty bucket ... I can switch back to CLI but well ... we won't be able to remove it then.

src/object_storage/s3.rs
benjaminch: Done.

test_utilities/src/aws.rs
benjaminch: Done

src/object_storage/s3.rs
deimosfr: Let's keep it this way

deimosfr (approved):

PR #571, opened 2022-01-23
Improve deployment time

evoxmusic:
Hi, here is a PR to improve the deployment time using Container Registry as a distributed container layers cache.

image

Note: this improvement works with all the cloud service providers and all the container registries (not only ECR)

### DIFF
--- a/src/build_platform/mod.rs
+++ b/src/build_platform/mod.rs
@@ -44,7 +44,7 @@ pub struct Build {
 }
 
 impl Build {
-    pub fn to_parent_build<P>(&self, clone_repo_into_dir: P) -> Result<Build, Error>
+    pub fn to_previous_build<P>(&self, clone_repo_into_dir: P) -> Result<Build, Error>
     where
         P: AsRef<Path>,
     {
--- a/src/build_platform/mod.rs
+++ b/src/build_platform/mod.rs
@@ -1,8 +1,13 @@
 use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
 
 use crate::error::{EngineError, EngineErrorCause, EngineErrorScope};
+use crate::git;
 use crate::models::{Context, Listen};
+use crate::utilities::get_image_tag;
+use git2::{Cred, CredentialType, Error};
 use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::path::Path;
 
 pub mod docker;
 pub mod local_docker;
@@ -16,6 +21,7 @@ pub trait BuildPlatform: Listen {
         format!("{} ({})", self.name(), self.id())
     }
     fn is_valid(&self) -> Result<(), EngineError>;
+    fn has_cache(&self, build: Build) -> Result<CacheResult, EngineError>;
     fn build(&self, build: Build, force_build: bool) -> Result<BuildResult, EngineError>;
     fn build_error(&self, build: Build) -> Result<BuildResult, EngineError>;
     fn engine_error_scope(&self) -> EngineErrorScope {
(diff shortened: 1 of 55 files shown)


src/container_registry/ecr.rs
```diff
@@ -263,6 +281,82 @@ impl ECR {
...
+                        self.name_with_id(),
+                    ),
+                ));
+            }
+        };
+
+        Ok(ECRCredentials::new(access_token, password, endpoint_url))
+    }
+
+    fn exec_docker_login(&self) -> Result<(), EngineError> {
```
benjaminch, line 332: Maybe worth moving this one into a common docker place. It's already used on a very similar way on SCW and DO already. Can be done later on though.

src/git.rs
```diff
@@ -46,7 +46,7 @@ fn authentication_callback<'a>(
     };
 }
 
-fn checkout(repo: &Repository, commit_id: &str) -> Result<(), Error> {
+fn checkout<'a>(repo: &'a Repository, commit_id: &'a str) -> Result<Object<'a>, Error> {
```
benjaminch, line 49: OPT: guts feeling but I guess we can avoid lifetime annotations here. Use String, use cloned Repo. I guess it doesn't add a lot of runtime perfs hit while simplifying the codebase avoiding future lifetime annotations leekage.

src/transaction.rs
```diff
@@ -134,6 +134,48 @@ impl<'a> Transaction<'a> {
...
+                );
+
+                return Err(err);
+            }
+        };
+
+        Ok(())
+    }
+
     fn _build_applications(
```
benjaminch, line 179: Same, why initial `_` ?

src/transaction.rs
```diff
@@ -134,6 +134,48 @@ impl<'a> Transaction<'a> {
         Ok(())
     }
 
+    fn _load_build_app_cache(&self, app: &crate::models::Application) -> Result<(), EngineError> {
```
benjaminch, line 137: why initial `_` ?

src/build_platform/mod.rs
```diff
@@ -37,20 +43,74 @@ pub struct Build {
     pub options: BuildOptions,
 }
 
+impl Build {
+    pub fn to_parent_build<P>(&self, clone_repo_into_dir: P) -> Result<Build, Error>
```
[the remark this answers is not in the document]
evoxmusic: `to_previous_build()` ? I use "parent" because git is a tree (parent node, child node)

src/container_registry/ecr.rs
evoxmusic: yes agreed

src/git.rs
evoxmusic: compiler complains if you remove the lifetime here

src/transaction.rs
evoxmusic: it's a personal convention to say that it is an internal function. Not important, we can remove it. (It's a convention used in Python). Happy to remove it.

src/transaction.rs
evoxmusic: same

src/transaction.rs
benjaminch: Ok, well, I guess the fact it's private is already a good signal it's internal. If you don't mind, removing it would help adding some coherence.

src/transaction.rs
benjaminch: 👍

src/build_platform/mod.rs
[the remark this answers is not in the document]
benjaminch: 👍

src/container_registry/ecr.rs
evoxmusic: Hum, after taking a look, I am sure it's done the same way for DockerHub, DOCR, and SCW Registry. For ECR we add the `endpoint URL` at the end of the `docker login` command. We can factorize it, but I wonder if it is really useful.

benjaminch: Ok !

tests/aws/aws_environment.rs
```diff
@@ -74,6 +79,97 @@ fn deploy_a_working_environment_with_no_router_on_aws_eks() {
...
+
+        let start_pull_time = SystemTime::now();
+        let _ = match ecr.pull(&image).unwrap() {
+            PullResult::Some(_) => assert!(true),
+            PullResult::None => assert!(false),
+        };
+
+        let pull_duration_2 = SystemTime::now().duration_since(start_pull_time).unwrap();
+
+        if pull_duration_2.as_millis() > pull_duration.as_millis() {
```
benjaminch, line 165: Hopefully I won't be flacky, also, maybe, once we got an idea of the % of speed up it brings, maybe worth adding it here. 
Because now, even if it's 1 milli quicker, it works, but maybe that's not what we want.

benjaminch (approved):
LGTM: just maybe we should think of a better way of testing the cache, not using only time, but something else, I have no idea at this point, but would be a nice addition :)

evoxmusic:
> 

Look, I also added some tests on the cached result. The time check is just to be sure that the pull with cache is faster than the one without it.

PR #572, opened 2022-01-24
fix(ENG-1068): normalize db service

MacLikorne:


lib/common/services/redis/templates/redis-master-svc.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "redis.fullname" . }}-master
+  name: {{ template "redis.fullname" . }}
```
erebe, line 5: is this our chart ? or one managed by bitnami ?
Because if it is the later, we need to mark somewhere to re-do this patch after an update of the chart;

erebe (approved):

lib/common/services/redis/templates/redis-master-svc.yaml
erebe: I guess, with storing a .patch file in our repository to know the diff we made
(i.e: with git patch)

erebe (dismissed):

benjaminch (approved):

PR #584, opened 2022-02-07
Fix deployment flow

erebe:
- Multiple stateful services trigger the on_create_check()
    before service are deployed every time

  - stateless services were check mulitples times also


benjaminch (approved):
LGTM

PR #607, opened 2022-02-15
Helm refacto

erebe:


src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -1390,12 +1359,12 @@ impl<'a> EKS<'a> {
                                 ),
                             ),
                             Err(e) => {
-                                let message_safe = format!("Error deleting chart `{}` deleted", chart.name);
+                                let message_safe = format!("Error deleting chart `{}` deleted: {}", chart.name, e);
```
deimosfr, line 1362: can you please rephrase

src/cloud_provider/helm.rs
```diff
@@ -108,6 +105,17 @@ impl ChartInfo {
         }
     }
 
+    pub fn new_from_release_name(name: &str, custom_namespace: &str) -> ChartInfo {
+        ChartInfo {
+            name: name.to_string(),
+            namespace: HelmChartNamespaces::Custom,
+            custom_namespace: Some(custom_namespace.to_string()),
+            timeout_in_seconds: 3600,
```
deimosfr, line 113: Why did you changed the default timeout? 3600 is far too high

src/cloud_provider/helm.rs
```diff
@@ -591,47 +587,34 @@ impl HelmChart for PrometheusOperatorConfigChart {
...
+                let prometheus_crds = [
+                    "prometheuses.monitoring.coreos.com",
+                    "prometheusrules.monitoring.coreos.com",
+                    "servicemonitors.monitoring.coreos.com",
+                    "podmonitors.monitoring.coreos.com",
+                    "alertmanagers.monitoring.coreos.com",
+                    "thanosrulers.monitoring.coreos.com",
+                ];
+                for crd in &prometheus_crds {
+                    let _ = kubectl_exec_delete_crd(kubernetes_config, crd, environment_variables.clone());
```
deimosfr, line 616: don't you manage errors?

src/cloud_provider/service.rs
```diff
@@ -365,30 +368,11 @@ pub fn deploy_user_stateless_service<T>(target: &DeploymentTarget, service: &T)
...
-                This issue often occurs due to ports misconfiguration. Make sure you exposed the correct port (using EXPOSE statement in Dockerfile or via Qovery configuration).",
-            ),
-            format!(
-                "{} {} has failed to start ⤬",
-                service.service_type().name(),
-                service.name_with_id()
-            ),
-        ),
-    )
+    deploy_stateless_service(target, service)
```
deimosfr, line 371: Did you forget to handle error?

src/cloud_provider/helm.rs
erebe: There was no default timeout before, I added this function for creating chartInfo from old HelmChart which wasn't containing any timeout.
After that it was using default max value for it, so no timeout ;'

src/cloud_provider/aws/kubernetes/mod.rs
erebe: sure

src/cloud_provider/service.rs
erebe: it is already the case, there is no ; so it return the result of deploy_stateless_service, which is the same type that the function retrun;
So it just propagate the error here

src/cloud_provider/helm.rs
erebe: We invoke it always now, so it returns an error if the crd is not present

src/cloud_provider/helm.rs
deimosfr: ..Default you're calling has a default one and 3600 is far too high

PR #608, opened 2022-02-15
refactor: extract state out of engine event

benjaminch:
This CL allows to remove state from engine event which will ease Logger
usage. Also, since state is carried by event details, there is no need
to have it at engine event level.

While it introduces info, warning and debug event, it flags former
states as deprecated. The latter will be removed in a future CL.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ extern crate tracing;
 extern crate trust_dns_resolver;
 
 pub mod build_platform;
+#[allow(deprecated)]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -4,6 +4,7 @@ extern crate tracing;
 extern crate trust_dns_resolver;
 
 pub mod build_platform;
+#[allow(deprecated)]
```
erebe, line 7: deprecated ?

erebe (approved):

PR #614, opened 2022-02-23
feat: handle charts with breaking changes

MacLikorne:


erebe:
Can you add a unit-test on the string split logic ?

erebe (approved):

PR #620, opened 2022-02-24
impr: reduce database tests duration

MacLikorne:


MacLikorne:
PR to reduce database tests duration and resources consumption.

- remove router for app connecting to db (no more domain check for the app)
- reduce replica for app connecting to db (from 2 to 1)

erebe (dismissed):

test_utilities/src/utilities.rs
```diff
@@ -417,12 +417,18 @@ pub fn generate_id() -> String {
     uuid
 }
 
-pub fn generate_password(allow_using_symbols: bool) -> String {
+pub fn generate_password(provider_kind: Kind, db_mode: DatabaseMode) -> String {
```
benjaminch, line 420: Not a big fan, to be check with others, but now this function heavily relies on Engine models, which makes this function harder to test eventually and will generate spaghettis code leaking logic everywhere. We should try to avoid leaking such logic and separate it as much as possible.

Also, another side effect is while this function is meant to generate a password for anything (could be DBs but also apps, or other secrets), now it's meant to DBs only, so name doesn't really match and so on.

So not a big fan on this one, should be handled somewhere else (in each DBs eventually since it's depend on DB type x cloud provider x Managed | Containered

PR #623, opened 2022-03-04
chore(ENG_1123_update_cert_manager): update chart manager

MacLikorne:


src/fs.rs
```diff
@@ -151,6 +154,130 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+}
+
+pub fn remove_lines_starting_with(path: String, starter: &str) -> Result<String, std::io::Error> {
+    info!("editing yaml backup file {}", path);
+
+    let mut file = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .open(path.as_str())
+        .expect("file doesn't exist");
```
erebe, line 205: return the error instead of panicing

src/fs.rs
```diff
@@ -151,6 +154,130 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+    let content = BufReader::new(file.try_clone().unwrap())
+        .lines()
+        .filter(|line| !line.as_ref().unwrap().contains(starter))
+        .map(|line| line.unwrap())
+        .collect::<Vec<String>>()
+        .join("\n");
+
+    file.flush()?;
+
+    file = OpenOptions::new()
```
erebe, line 216: you open the file twice

src/fs.rs
```diff
@@ -151,6 +154,130 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+        .expect("file doesn't exist");
+
+    let content = BufReader::new(file.try_clone().unwrap())
+        .lines()
+        .filter(|line| !line.as_ref().unwrap().contains(starter))
+        .map(|line| line.unwrap())
+        .collect::<Vec<String>>()
+        .join("\n");
+
+    file.flush()?;
```
erebe, line 214: no need to flush, you are only reading the file. there is nothing to commit/flush to disk

src/fs.rs
```diff
@@ -151,6 +154,130 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+    let files = WalkDir::new(working_root_dir)
+        .follow_links(true)
+        .into_iter()
+        .filter_map(|e| e.ok());
+    let mut backup_paths: Vec<String> = vec![];
+    for file in files {
+        if file
+            .file_name()
+            .to_str()
+            .expect("No file name.")
```
erebe, line 242: same, expect will make the engine panic/crash if it is not valid.
return a proper Io::Error

src/cmd/helm.rs
```diff
@@ -524,6 +526,105 @@ impl Helm {
 
         Ok(())
     }
+
+    pub fn prepare_chart_backup<P>(
```
erebe, line 530: Move this outside of the helm module/crate.
It is not part of the functionality of the helm command.

Just create a helper function that does this for you, outside of this module.

Cargo.toml
```diff
@@ -55,6 +55,7 @@ tera = "1.10.0"
 serde = "1.0.126"
 serde_json = "1.0.64"
 serde_derive = "1.0.126"
+serde_yaml = "0.8.23"
```
benjaminch: Can be moved to [dev-dependencies]

src/cmd/helm.rs
```diff
@@ -492,6 +496,21 @@ impl Helm {
 
         let mut error_message: Vec<String> = vec![];
 
+        let tmp_dir = std::env::temp_dir();
+        let root_dir_path = Path::new(tmp_dir.as_path());
+
+        if chart.backup_resources.is_some() {
+            match self.prepare_chart_backup(
```
benjaminch, line 503: OPT: can you if let for this one since you do return only on Err

src/cmd/helm.rs
```diff
@@ -506,7 +525,15 @@ impl Helm {
 
         match helm_ret {
             // Ok is ok
-            Ok(_) => Ok(()),
+            Ok(_) => {
+                if chart.backup_resources.is_some() {
+                    match self.apply_chart_backup(root_dir_path, &self.get_all_envs(envs), chart) {
```
benjaminch, line 530: OPT: same here, if let can be used

src/cmd/helm.rs
```diff
@@ -561,6 +588,150 @@ impl Helm {
...
+        let mut backups: Vec<(String, String)> = vec![];
+        for backup_resource in backup_resources {
+            match kubectl_get_resource_yaml(
+                &self.kubernetes_config,
+                envs.to_vec(),
+                backup_resource.as_str(),
+                Some(chart.get_namespace_string().as_str()),
+            ) {
+                Ok(content) => {
+                    if !content.contains("No resources found") {
```
benjaminch, line 611: OPT: Maybe make sure to compare strings in lower case just to be sure

src/cmd/helm.rs
```diff
@@ -561,6 +588,150 @@ impl Helm {
...
+
+        if backups.len() == 0 {
+            return Ok(backup_infos);
+        }
+
+        for backup in backups.clone() {
+            match create_yaml_backup_file(
+                workspace_root_dir.as_ref(),
+                chart.name.to_string(),
+                Some(backup.0.clone()),
```
benjaminch, line 631: OPT: Not very clear at first read. Maybe worth using a struct avoiding to have to index like this.

src/cmd/helm.rs
```diff
@@ -561,6 +588,150 @@ impl Helm {
...
+                backup.1,
+            ) {
+                Ok(path) => {
+                    backup_infos.push((backup.0, path));
+                }
+                Err(e) => {
+                    return Err(CmdError(
+                        chart.name.clone(),
+                        HelmCommand::UPGRADE,
+                        CommandError::new(e.to_string(), Some(e.to_string()), None),
```
benjaminch, line 641: Might lead to some leak here, are you sure error message is safe and doesn't contains any secrets / sensitive data?
Please go with:
```
CommandError::new("Error while trying to create YAML backup file".to_string(), Some(e.to_string()), None)
```

Even better, maybe you can have `create_yaml_backup_file` to return a `CommandError` directly so it handles safe and unsafe message for callers.

src/cmd/helm.rs
```diff
@@ -561,6 +588,150 @@ impl Helm {
...
+                }
+            }
+        }
+
+        for backup_info in backup_infos.clone() {
+            if let Err(e) = remove_lines_starting_with(backup_info.1.clone(), "resourceVersion") {
+                return Err(CmdError(
+                    chart.name.clone(),
+                    HelmCommand::UPGRADE,
+                    CommandError::new(e.to_string(), Some(e.to_string()), None),
```
benjaminch, line 652: Same, CF previous

Might lead to some leak here, are you sure error message is safe and doesn't contains any secrets / sensitive data?
Please go with:
```
CommandError::new("Error while trying to create YAML backup file".to_string(), Some(e.to_string()), None)
```

Even better, maybe you can have `create_yaml_backup_file` to return a `CommandError` directly so it handles safe and unsafe message for callers.

src/cmd/helm.rs
```diff
@@ -561,6 +588,150 @@ impl Helm {
...
+                ));
+            }
+
+            if let Err(e) = remove_lines_starting_with(backup_info.1.clone(), "uid") {
+                return Err(CmdError(
+                    chart.name.clone(),
+                    HelmCommand::UPGRADE,
+                    CommandError::new(e.to_string(), Some(e.to_string()), None),
```
benjaminch, line 660: Same, CF previous

Might lead to some leak here, are you sure error message is safe and doesn't contains any secrets / sensitive data?
Please go with:
```
CommandError::new("Error while trying to create YAML backup file".to_string(), Some(e.to_string()), None)
```

Even better, maybe you can have `create_yaml_backup_file` to return a `CommandError` directly so it handles safe and unsafe message for callers.

src/fs.rs
```diff
@@ -151,6 +155,141 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
     }
 }
 
+pub fn create_yaml_backup_file<P>(
+    working_root_dir: P,
+    chart_name: String,
+    resource_name: Option<String>,
+    content: String,
+) -> Result<String, std::io::Error>
```
benjaminch, line 163: OPT: Maybe return a CommandError here, so this function takes care of sending back safe and unsafe message properly.

src/fs.rs
```diff
@@ -151,6 +155,141 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+            file.flush()?;
+            Ok(path
+                .to_str()
+                .map(|e| e.to_string())
+                .ok_or_else(|| Error::from(ErrorKind::NotFound))?)
+        }
+    }
+}
+
+pub fn remove_lines_starting_with(path: String, starter: &str) -> Result<String, std::io::Error> {
```
benjaminch, line 214: OPT: Maybe return a CommandError here, so this function takes care of sending back safe and unsafe message properly.

src/fs.rs
```diff
@@ -151,6 +155,141 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+    info!("editing yaml backup file {}", path);
```
benjaminch, line 215: OPT: Not big fan of such logs, maybe remove

src/fs.rs
```diff
@@ -151,6 +155,141 @@ pub fn create_workspace_archive(working_root_dir: &str, execution_id: &str) -> R
...
+    };
+
+    let content = match decode(secret_content) {
+        Ok(bytes) => from_utf8_lossy(&bytes[1..bytes.len() - 1]).to_string(),
+        Err(_) => return Err(CommandError::new(message.clone(), Some(message), None)),
+    };
+    match create_yaml_backup_file(working_root_dir.as_ref(), secret.metadata.name.clone(), None, content) {
+        Ok(path) => Ok(path),
+        Err(e) => {
+            let message = format!("Unable to create backup file from secret {}: {}", secret.metadata.name, e);
```
benjaminch, line 287: Please don't do that, it can leak stuff. Make sure safe message is safe and do not put error message directly into the safe message.

```
Err(CommandError::new(
  format!("Unable to create backup file from secret {}.", secret.metadata.name), 
  Some(e.to_string()),
  None))
```

benjaminch (commented):
Some error handling can lead to leakage, please make sure everything sent to `CommandError` safe message is actually safe and provide full error details into unsafe part of `CommandError`.

benjaminch (approved):

src/cmd/helm.rs
```diff
@@ -506,7 +521,16 @@ impl Helm {
 
         match helm_ret {
             // Ok is ok
-            Ok(_) => Ok(()),
+            Ok(_) => {
+                if chart.backup_resources.is_some() {
+                    if let Err(e) =
+                        apply_chart_backup(&self.kubernetes_config, &root_dir_path, &self.get_all_envs(envs), chart)
```
erebe, line 527: What happens to the backup if the upgrade fails ?

src/cmd/helm.rs
```diff
@@ -58,7 +59,7 @@ pub enum HelmError {
 
 #[derive(Debug)]
 pub struct Helm {
-    kubernetes_config: PathBuf,
+    pub kubernetes_config: PathBuf,
```
erebe, line 62: is the pub still necessary ? it seems fishy

MacLikorne: forgot to remove it after file architecture changed.

src/cmd/helm.rs
MacLikorne: nothing, it should be deleted I think

erebe: where ?

src/cmd/helm.rs
```diff
@@ -533,6 +546,9 @@ impl Helm {
             }
             Err(err) => {
                 error!("Helm error: {:?}", err);
+                if let Err(e) = delete_unused_chart_backup(&self.kubernetes_config, &self.get_all_envs(envs), chart) {
+                    return Err(e);
```
erebe, line 550: - You don't check if need_backup was true, if it was not the get_secret in the delete is going to fail

- You return the error of to delete_unsued, while we are interested in the upgrade error, it is going to be confusing in log/to troubleshoot.

I was not very found of putting the logic of backup/restore in the helm upgrade at first, because this is nothing to with the helm command at all.
But the more you add logic of this in this function, you more we see that it tries to mix 2 different things in one.

Just extract the logic of backup/restore outside this helm function and call it above

if Err(err) = helm_utils::backup_ressources(xxxx) {
     return Err(err);
}

match helm::upgrade(xxxx) {
 Ok(x) =>  {
    helm_utils::restore(xxxx); // rollback helm if it fails ?
    Ok(x)
 } 
 Err(x) { helm_utils::delete_backup(); xxxx } 
}

it should be your prepare_backup function that should do the check of versions

erebe (approved):

PR #629, opened 2022-03-07
Improve docker build time

mzottola:


src/container_registry/docker.rs
```diff
@@ -189,6 +190,59 @@ pub fn docker_tag_and_push_image(
             info!("image {} has successfully been pushed", image_with_tag);
             Ok(())
         }
+    };
+
+    let image_with_latest_tag = format!("{}:latest", image_name);
```
erebe, line 195: re-use name_with_latest_tag()

erebe (dismissed):

src/container_registry/docker.rs
mzottola: done

src/transaction.rs
```diff
@@ -125,44 +125,28 @@ impl<'a> Transaction<'a> {
...
+                        "something goes wrong while pulling image from {:?} container registry",
+                        container_registry.kind()
+                    ))
+                );
+                return Err(EngineError::new_from_legacy_engine_error(err));
             }
         };
 
-        Ok(())
+        return Ok(());
```
benjaminch, line 149: OPT: no need for explicit `return` can be `Ok(())` without `;`

benjaminch (dismissed):

src/transaction.rs
mzottola: done

benjaminch (approved):

erebe (approved):

PR #635, opened 2022-03-08
Add tag 'latest' for buildpack images as well

mzottola:


mzottola:
tested with buildpack v0.24

benjaminch (approved):

PR #638, opened 2022-03-11
feat: migrate container registries to new logging

benjaminch:
Ticket: ENG-1136

### DIFF
--- a/src/cmd/command.rs
+++ b/src/cmd/command.rs
@@ -32,6 +32,19 @@ pub enum CommandError {
     Killed(String),
 }
 
+impl CommandError {
+    pub fn to_string(&self) -> String {
+        match self {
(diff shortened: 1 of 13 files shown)


erebe (approved):

MacLikorne (approved):

src/cmd/command.rs
```diff
@@ -32,6 +32,19 @@ pub enum CommandError {
     Killed(String),
 }
 
+impl CommandError {
```
erebe, line 35: thisError already implement Display trait.
is there a need to add a to_string() ?

benjaminch: You are right, here's a MR to change that => https://github.com/Qovery/engine/pull/640

PR #654, opened 2022-03-21
Refacto to having double compute of image hash

erebe:
In summary, all this PR does is avoid recreating multiple time the same object in order to be able to mutate the image hash and have it correctly set later on in the stage

### DIFF
--- a/src/build_platform/mod.rs
+++ b/src/build_platform/mod.rs
@@ -20,7 +20,7 @@ pub trait BuildPlatform: ToTransmitter + Listen {
         format!("{} ({})", self.name(), self.id())
     }
     fn is_valid(&self) -> Result<(), EngineError>;
-    fn build(&self, build: Build, is_task_canceled: &dyn Fn() -> bool) -> Result<BuildResult, EngineError>;
+    fn build(&self, build: &Build, is_task_canceled: &dyn Fn() -> bool) -> Result<(), EngineError>;
     fn logger(&self) -> Box<dyn Logger>;
     fn get_event_details(&self) -> EventDetails {
         let context = self.context();
@@ -94,6 +94,9 @@ impl Image {
         self.registry_url.host_str().unwrap()
     }
 
+    pub fn repository_name(&self) -> &str {
+        self.name.split('/').collect::<Vec<&str>>()[0]
+    }
     pub fn full_image_name_with_tag(&self) -> String {
         format!(
             "{}/{}:{}",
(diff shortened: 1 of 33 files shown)


src/build_platform/mod.rs
```diff
@@ -94,6 +94,9 @@ impl Image {
         self.registry_url.host_str().unwrap()
     }
 
+    pub fn repository_name(&self) -> &str {
+        self.name.split('/').collect::<Vec<&str>>()[0]
```
benjaminch, line 98: What about computing this field only once at Image creation?

src/cloud_provider/aws/application.rs
```diff
@@ -20,7 +20,7 @@ use crate::logger::Logger;
 use crate::models::{Context, Listen, Listener, Listeners, ListenersHelper, Port};
 use ::function_name::named;
 
-pub struct Application {
+pub struct ApplicationAws {
```
benjaminch, line 23: Out of curiosity, what's the reason adding provider prefix?

src/cloud_provider/aws/databases/mongodb.rs
```diff
@@ -92,13 +92,17 @@ impl MongoDB {
     }
 }
 
-impl StatefulService for MongoDB {
+impl StatefulService for MongoDbAws {
+    fn as_stateful_service(&self) -> &dyn StatefulService {
```
benjaminch, line 96: OPT: maybe use `impl From<MongoDbAws> for StatefulService` ? I think you should add `From<MongoDbAws>` to `StatefulService` trait declaration.

src/cloud_provider/aws/application.rs
erebe: To avoid confusion, having 4 times the same struct (1 for each provider + 1 for the trait) make it really hard to know where you are when just looking at the code.
It avoids also the silly conflicting imports

src/cloud_provider/aws/databases/mongodb.rs
erebe: You can't because when you are in the From<xxxx> the compiler don't know that MongoDbAws is also a StatefulService
Rust does not support upcasting (yet) https://stackoverflow.com/questions/28632968/why-doesnt-rust-support-trait-object-upcasting

src/build_platform/mod.rs
erebe: done

benjaminch (approved):

PR #661, opened 2022-03-24
Skip build if image already exist remotly

erebe:


src/build_platform/local_docker.rs
```diff
@@ -108,6 +108,25 @@ impl LocalDocker {
         lh: &ListenersHelper,
         is_task_canceled: &dyn Fn() -> bool,
     ) -> Result<(), BuildError> {
+        // logger
+        let log_info = {
+            let app_id = build.image.application_id.clone();
+            move |msg: String| {
+                info!("{}", msg);
```
benjaminch, line 115: No need, `logger.log` does it.

PR #669, opened 2022-03-28
Refacto application to avoid duplication

erebe:
Make application a generic struct that can be parametrized by a cloud provider kind.

This is to avoid duplication of struct and behavior.
The only specific stuff is tera_context (that can be externalized in a specific trait later to avoid re-implementing the whole service trait)


src/models/application.rs
```diff
@@ -0,0 +1,379 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum ApplicationError {
+    #[error("Application invalid configuration: {0}")]
+    InvalidConfig(String),
+}
+
+pub struct Application<T: CloudProvider> {
+    _marker: PhantomData<T>,
+    pub(crate) context: Context,
```
benjaminch, line 29: OPT: Can we avoid exposing internal fields to the outside world, making those private by default?

src/models/digital_ocean/application.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        if self.context.resource_expiration_in_seconds().is_some() {
+            context.insert("resource_expiration_in_seconds", &self.context.resource_expiration_in_seconds())
+        }
+
+        Ok(context)
+    }
+}
+
+impl Service for Application<DO> {
```
MacLikorne, line 95: Why don't you implement Service directly at Application declaration (models/application.rs) ?

src/models/application.rs
erebe: it is pub(crate) so it is only exposed inside the models/ directory/crate, not outside

src/models/digital_ocean/application.rs
erebe: Due to tera_context function, it is specific per provider.
Should a add specific trait to only implement this method instead of the whole Service trait

src/models/application.rs
benjaminch: Sure, but why need to expose it even to models/ directory/crate?

erebe: So that each cloud provider can use them, if you are in the implementation Application<AWS> if it is not in pub(crate) you can't access self.port for example

benjaminch: ok ok

benjaminch (approved):

PR #670, opened 2022-03-29
Router refacto preparation advanced settings

erebe:


src/cloud_provider/service.rs
```diff
@@ -157,7 +157,7 @@ pub trait StatefulService: Service + Create + Pause + Delete {
     fn is_managed_service(&self) -> bool;
 }
 
-pub trait Router: StatelessService + Listen + Helm {
+pub trait IRouter: StatelessService + Listen + Helm {
```
benjaminch, line 160: CS: I would advise to find another name, `IXxxx` sounds too much C# or Java style.
If `Router` is not available, then maybe `ServiceRouter` or anything like so.

erebe: What does mean CS ?
Well, don't mind doing it with next PR.

But honestly Ixxx or xxxService, it is kind of the same to me, as long as we can easily distinguish between the trait and the struct. 
But going to rename if you want

benjaminch: CS stands for coding style sorry,

IMO, `IRouter` is understandable but is not consitent with what we have in place right now in the code base nor with rust naming convention https://rust-lang.github.io/rfcs/0344-conventions-galore.html#trait-naming 

What's wrong with initial `Router` name? :)

erebe: It conflicts when you want to import the 2 of them, so you have to fully qualified the name of one type, which is further ugly.
Also, it is hard to distinguish if the function is taking the struct or the trait, which is important in some part of the code, as it allow to see where in the layers we are/should be.

It let you pick the name you want, just le me know in comment, going to rename it after this PR.

benjaminch: as discussed offline, let's use `RouterImpl ` for implementations and `Router` for trait

benjaminch (approved):

PR #674, opened 2022-03-30
Refacto AWS Databases

erebe:


src/cloud_provider/service.rs
```diff
@@ -250,32 +251,32 @@ pub struct DatabaseOptions {
...
-    Redis(&'a DatabaseOptions),
+pub enum DatabaseType {
+    PostgreSQL,
+    MongoDB,
+    MySQL,
+    Redis,
 }
 
-impl<'a> ToString for DatabaseType<'a> {
+impl ToString for DatabaseType {
```
benjaminch, line 261: OPT: I know it's not part of this change, might be a FAR

Usually better to implement Display for enum / struct.
We should not implement [ToString](http://doc.rust-lang.org/std/string/trait.ToString.html) manually. `ToString` trait is already implemented for all types which implement `fmt::Display`.

Implementing `Display`, `to_string()` will be available on struct / enum automatically.

src/models/aws/database.rs
```diff
@@ -0,0 +1,348 @@
...
+};
+use crate::models::database_utils::{
+    get_self_hosted_mongodb_version, get_self_hosted_mysql_version, get_self_hosted_postgres_version,
+    get_self_hosted_redis_version,
+};
+use crate::models::types::{ToTeraContext, AWS};
+use tera::Context as TeraContext;
+
+/////////////////////////////////////////////////////////////////
+// MANAGED
```
benjaminch, line 23: Shouldn't it be `CONTAINER`?

src/cloud_provider/service.rs
erebe: yeah, I saw it too, but I let that for another time :;(

benjaminch (approved):

PR #683, opened 2022-04-08
fix: desired_nodes is now redefined after a cluster resize

benamib:


src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -547,6 +560,106 @@ impl EKS {
         Ok(context)
     }
 
+    // Return a pair of:
```
benjaminch, line 563: OPT: use Rust doc style here 
https://doc.rust-lang.org/rust-by-example/meta/doc.html

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -547,6 +560,106 @@ impl EKS {
...
+                    event_details,
+                    Option::Some(CommandError::new_from_safe_message(
+                        "Could not find node_group in terra context".to_string(),
+                    )),
+                ))
+            }
+        };
+
+        let region = RusotoRegion::from_str(&self.region())
+            .unwrap_or_else(|_| panic!("S3 region `{}` doesn't seems to be valid.", self.region.to_aws_format()));
```
benjaminch, line 580: Should not fire a panic here, please return an Engine Error

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -547,6 +560,112 @@ impl EKS {
         Ok(context)
     }
 
+    /// Returns a tuple of (update_desired_node: bool, desired_nodes_count: i32).
+    fn should_update_desired_nodes(&self, event_details: EventDetails) -> Result<(bool, i32), EngineError> {
+        let future_node_group = match self.nodes_groups.is_empty() {
+            false => self.nodes_groups.first().unwrap(),
```
deimosfr, line 566: Please avoid unwrap in prod, use expect instead, will be easier to debug (even if in this case it should never be a problem)

tests/aws/aws_whole_enchilada.rs
```diff
@@ -51,3 +51,48 @@ fn create_upgrade_and_destroy_eks_cluster_with_env_in_eu_west_3() {
...
+            Kind::Aws,
+            context.clone(),
+            logger(),
+            region,
+            Some(aws_zones),
+            ClusterTestType::WithNodesResize,
+            AWS_KUBERNETES_MAJOR_VERSION,
+            AWS_KUBERNETES_MINOR_VERSION,
+            &ClusterDomain::Custom(cluster_domain),
+            Some(WithNatGateways),
```
MacLikorne, line 94: You don't need NatGateway for this test, it will make it longer and more expensive. And you don't need env neither. You're testing node resizing, no need to test the rest, it's already done.

tests/aws/aws_whole_enchilada.rs
```diff
@@ -51,3 +51,48 @@ fn create_upgrade_and_destroy_eks_cluster_with_env_in_eu_west_3() {
...
+        "{}.{}",
+        cluster_id.as_str(),
+        secrets
+            .DEFAULT_TEST_DOMAIN
+            .as_ref()
+            .expect("DEFAULT_TEST_DOMAIN is not set in secrets")
+            .as_str()
+    );
+
+    let environment = test_utilities::common::working_minimal_environment(&context, cluster_domain.as_str());
```
MacLikorne, line 79: you don't need those I think

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -547,6 +560,112 @@ impl EKS {
...
+            return Err(EngineError::new_cluster_has_no_worker_nodes(
+                event_details,
+                Option::Some(CommandError::new_from_safe_message(format!(
+                    "Could not find node_groups for cluster {}",
+                    self.cluster_name()
+                ))),
+            ));
+        }
+
+        let actual_nodes_group = match block_on(eks_client.describe_nodegroup(DescribeNodegroupRequest {
```
deimosfr, line 614: Don't you have a list? Today we only have one, but technically we can have several. May be having another way (filters) to distinguish them could be great

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -547,6 +560,112 @@ impl EKS {
...
+
+        let should_update_desired_nodes = scaling_config.min_size.unwrap_or_default()
+            != i64::from(future_node_group.min_nodes)
+            || scaling_config.max_size.unwrap_or_default() != i64::from(future_node_group.max_nodes);
+
+        let kubeconfig = match self.get_kubeconfig_file() {
+            Ok((path, _)) => path,
+            Err(_) => return Ok((false, scaling_config.desired_size.unwrap() as i32)),
+        };
+        let get_node_result = retry::retry(Fixed::from_millis(10000).take(60), || {
```
deimosfr, line 640: Too huge

deimosfr: yep

benamib: I have changed it to 5, is this better ?

PR #684, opened 2022-04-11
Cleanup docker image purge

erebe:


src/build_platform/local_docker.rs
```diff
@@ -70,32 +72,48 @@ impl LocalDocker {
...
+            "Purging docker images to reclaim disk space. Only {} % disk free space, This may take some time",
+            disk_free_space_percent
+        );
+        self.logger
+            .log(EngineEvent::Info(self.get_event_details(), EventMessage::new_from_safe(msg)));
+
+        // Request a purge if a disk is being low on space
+        if let Err(err) = self.context.docker.prune_images() {
+            let msg = format!("Error while purging docker images: {}", err);
+            error!("{}", &msg);
```
benjaminch, line 114: no need, can be removed

erebe: I want to put it in error in the engine logs ^^

benjaminch: It's done on the next line already, using `logger.log` :)

benjaminch: I guess, we can just add a function to log error from a string, it would be a nice add I guess :)

PR #687, opened 2022-04-13
refactor: update docker disk purger percentage

deimosfr:


mzottola (approved):

src/build_platform/local_docker.rs
```diff
@@ -73,7 +73,7 @@ impl LocalDocker {
         }
 
         // arbitrary percentage that should make the job anytime
-        const DISK_FREE_SPACE_PERCENTAGE_BEFORE_PURGE: u64 = 20;
+        const DISK_FREE_SPACE_PERCENTAGE_BEFORE_PURGE: u64 = 40;
```
evoxmusic, line 76: u64 -> u8

MacLikorne (approved):

src/build_platform/local_docker.rs
deimosfr: just to avoid casting on other functions from libs returning u64

PR #688, opened 2022-04-14
feat: speed up env delete

deimosfr:
Speed up environment delete and fix possible issues occuring because of required redeploy before destroy terraform side

### DIFF
--- a/src/cloud_provider/kubernetes.rs
+++ b/src/cloud_provider/kubernetes.rs
@@ -22,7 +22,8 @@ use crate::cloud_provider::{service, CloudProvider, DeploymentTarget};
 use crate::cmd::kubectl;
 use crate::cmd::kubectl::{
     kubectl_delete_objects_in_all_namespaces, kubectl_exec_count_all_objects, kubectl_exec_delete_pod,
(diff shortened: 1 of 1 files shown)


src/cloud_provider/kubernetes.rs
```diff
@@ -668,6 +669,19 @@ pub fn delete_environment(
...
+    let kubeconfig = kubernetes.get_kubeconfig_file_path()?;
+
+    // check if environment is not already deleted
+    // speed up delete env because of terraform requiring apply + destroy
+    if !kubectl_exec_is_namespace_present(
+        kubeconfig.clone(),
+        environment.namespace(),
+        kubernetes.cloud_provider().credentials_environment_variables(),
+    ) {
+        info!("no need to delete environment {}, already absent", environment.namespace());
```
benjaminch, line 681: Please avoid using tracing lib when you want to log something on core side and locally with all the context.
Here you've got logger and event_details provided in the function so you can use it:

logger.log(EngineEvent::Info(event_details, EventMessage::from_safe_message("your log"));

PR #689, opened 2022-04-15
feat: add application commit ID to logs

benjaminch:
This CL adds:
- application commit ID to some application actions logs.
- application commit ID annotation / label to application deployment
  yaml so it can be retrieved later on from other services

Ticket: ENG-1170


erebe:
ok, but not really sure why you want to add it in the deployment logs.
It will be always the app commit id that have been  configured.
It is only in the application logs when you do a rollout that you can get multiple version at the same time and that it can be confusing for the user

erebe (dismissed):

PR #694, opened 2022-04-20
feat: Qovery AWS EC2 🌎

deimosfr:


test_utilities/src/common.rs
```diff
@@ -1135,10 +1136,11 @@ pub fn test_db(
         Kind::Aws => (AWS_TEST_REGION.to_string(), AWS_KUBERNETES_VERSION.to_string()),
         Kind::Do => (DO_TEST_REGION.to_string(), DO_KUBERNETES_VERSION.to_string()),
         Kind::Scw => (SCW_TEST_ZONE.to_string(), SCW_KUBERNETES_VERSION.to_string()),
+        Kind::Edge(Edge::Aws) => (AWS_TEST_REGION.to_string(), AWS_K3S_VERSION.to_string()),
```
erebe, line 1139: Can you flatten the kind to `AWS_EDGE` instead of nesting 2 types inside.
Even if we extend that to other providers, there is little chances that we want to first match on Edge before willing to know which provider it is

evoxmusic: I've redefined this part. I think `Qovery Edge` is a misleading name and does not reflect the goal of this feature (deploying Qovery on an AWS EC2 instance). What I propose is to just create a new Kubernetes provider named `EC2` implementing the `Kubernetes` trait. So we do not need to touch anything. We just need to implement EC2 as a Kubernetes type. Then, as a user, you can choose between EKS and EC2 while creating a cluster on Qovery.

benjaminch:
I find `bootstrap` lib folder a bit complicated now. 
Shouldn't we treat AWS EC2 / Edge as a new provider? I will bring a lot of upsides:
- Easier to maintain (even if in a first time code will be duplicated a bit, it's not an issue)
- Easier to unplug eventually at some point if needed
- Easier to understand rather than having mixed files everywhere

We can reuse stuff from one provider to the other, or at least make the change easily to support it (I am thinking about applications / DBs). New AWS EC2 provider should be able to reuse full fledge components as we do today for DNS managers / container registries or Object storages. It will be nice to allow that.

So for 'lib/bootstrap' I would go with something like:
```
/lib
  /aws
    /bootstrap
    /charts
    /chart-values
    /[..]
  /aws-ec2 <= Only AWS EC2 specifics goes here, if same as AWS, then just copy for now, we will refactor if needed at some point
    /bootstrap
    /charts
    /chart-values
    /[..]
  /scaleway
    /bootstrap
    /charts
    /chart-values
    /[..]
  /digitalocean
    /bootstrap
    /charts
    /chart-values
    /[..]
```

evoxmusic:
> 

EC2 is not a new `Cloud Provider` but a new `Kubernetes Provider`. So everything that is AWS-related is valid to this new Kubernetes Provider. 

⚠️ Note: the name `Qovery Edge` has been replaced by `Qovery AWS EC2` since it is not an Edge provider that we really want here. Our use case is to make Qovery accessible on AWS with a cost-effective solution while keeping Kubernetes.

benjaminch:
Sure thing but I think we should consider this new EC2 as new provider or at least better isolate / separate it in the code base while avoiding to pollute general AWS cases. Of course there will be some duplicate but it can be refactored easily.

evoxmusic:
I've refactored the code to not duplicate them between EKS and EC2 - if you want we can move EC2 and EKS into separate files. It makes sense to me as well. Let me know

benjaminch:
It would makes sense to split those and see how we can refactor a bit after. It can be done after but let say doing it now is less work upfront I think. Let me know if I can help.

evoxmusic:
> It would makes sense to split those and see how we can refactor a bit after. It can be done after but let say doing it now is less work upfront I think. Let me know if I can help.

@benjaminch I did the refacto :) Let me know what you think now. Thx for your feedback

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+        &listeners_helper,
+    );
 
-        self.logger().log(EngineEvent::Info(
-            self.get_event_details(Stage::Infrastructure(InfrastructureStep::Pause)),
-            EventMessage::new_from_safe("Preparing EKS cluster pause.".to_string()),
-        ));
+    kubernetes.logger().log(EngineEvent::Info(
+        kubernetes.get_event_details(Stage::Infrastructure(InfrastructureStep::Pause)),
+        EventMessage::new_from_safe(format!("Preparing {} cluster pause.", kubernetes.kind())),
```
benjaminch, line 777: OPT: `kubernetes.kind()` shouldn't be needed here. Logging system does add this info, no need to add it.

 `EventMessage::new_from_safe("Preparing cluster pause.".to_string())`
 
 Should do the trick.

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+            kubernetes.name(),
+            kubernetes.id()
+        )
+        .as_str(),
+        &listeners_helper,
+    );
+
+    kubernetes.logger().log(EngineEvent::Info(
+        event_details.clone(),
+        EventMessage::new_from_safe("Pausing EKS cluster deployment.".to_string()),
```
benjaminch, line 914: OPT: Maybe a miss? 

Again, `EventMessage::new_from_safe("Pausing cluster deployment.".to_string())` should be enough

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+    ));
+
+    match terraform_exec(temp_dir.as_str(), terraform_args) {
+        Ok(_) => {
+            let message = format!("Kubernetes cluster {} successfully paused", kubernetes.name());
```
benjaminch, line 919: FAR: note for later, but now on, cluster ID, cluster name and co are populated automatically into the log, no need to add it.

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+            "Deleting all non-Qovery deployed applications and dependencies for cluster {}/{}",
+            kubernetes.name(),
+            kubernetes.id()
         );
-        self.send_to_customer(&message, &listeners_helper);
-        self.logger()
-            .log(EngineEvent::Info(event_details.clone(), EventMessage::new_from_safe(message)));
 
-        self.logger().log(EngineEvent::Info(
+        kubernetes.logger().log(EngineEvent::Info(
```
benjaminch, line 1043: OPT: no need to add extra info:

```
kubernetes.logger().log(EngineEvent::Info(
            event_details.clone(),
            EventMessage::new_from_safe("Deleting all non-Qovery deployed applications and dependencies for cluster".to_string()),
        ));
```

In this case I think it's not important since this message is reused to be sent to old logging system, but just to keep in mind.

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+            for chart in charts_to_delete {
+                let chart_info = ChartInfo::new_from_release_name(&chart.name, &chart.namespace);
+                match helm.uninstall(&chart_info, &[]) {
+                    Ok(_) => kubernetes.logger().log(EngineEvent::Info(
+                        event_details.clone(),
+                        EventMessage::new_from_safe(format!("Chart `{}` deleted", chart.name)),
+                    )),
                     Err(e) => {
-                        return Err(EngineError::new_terraform_error_while_executing_pipeline(event_details, e));
+                        let message_safe = format!("Can't delete chart `{}`: {}", &chart.name, e);
```
benjaminch, line 1153: Potential leak here, please do not fill error message into safe message:

```
kubernetes.logger().log(EngineEvent::Warning(
                            event_details.clone(),
                            EventMessage::new(format!("Can't delete chart `{}`", &chart.name), Some(e.to_string())),
                        ))
```

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+            Ok(helm_charts) => {
+                for chart in helm_charts {
+                    let chart_info = ChartInfo::new_from_release_name(&chart.name, &chart.namespace);
+                    match helm.uninstall(&chart_info, &[]) {
+                        Ok(_) => kubernetes.logger().log(EngineEvent::Info(
+                            event_details.clone(),
+                            EventMessage::new_from_safe(format!("Chart `{}` deleted", chart.name)),
+                        )),
+                        Err(e) => {
+                            let message_safe = format!("Error deleting chart `{}`: {}", chart.name, e);
```
benjaminch, line 1205: Same, potential leak here, do not fill error message into safe message.

```
kubernetes.logger().log(EngineEvent::Warning(
                                event_details.clone(),
                                EventMessage::new(format!("Error deleting chart `{}`", chart.name), Some(e.to_string())),
                            ))
```

test_utilities/src/common.rs
```diff
@@ -13,18 +13,22 @@ use qovery_engine::io_models::{
...
-use qovery_engine::cloud_provider::aws::kubernetes::{VpcQoveryNetworkMode, EKS};
+use qovery_engine::cloud_provider::aws::kubernetes::ec2::EC2;
+use qovery_engine::cloud_provider::aws::kubernetes::eks::EKS;
+use qovery_engine::cloud_provider::aws::kubernetes::VpcQoveryNetworkMode;
 use qovery_engine::cloud_provider::aws::regions::{AwsRegion, AwsZones};
 use qovery_engine::cloud_provider::aws::AWS;
 use qovery_engine::cloud_provider::digitalocean::kubernetes::DOKS;
 use qovery_engine::cloud_provider::digitalocean::DO;
 use qovery_engine::cloud_provider::environment::Environment;
+use qovery_engine::cloud_provider::kubernetes::Kind as KKind;
```
benjaminch, line 31: OPT: rename to a clearer alias maybe? `K8sKind` or `KubernetesKind`

tests/aws/aws_kubernetes_ec2.rs
```diff
@@ -0,0 +1,68 @@
...
+use qovery_engine::cloud_provider::aws::kubernetes::VpcQoveryNetworkMode::{WithNatGateways, WithoutNatGateways};
+use qovery_engine::cloud_provider::aws::regions::AwsRegion;
+use qovery_engine::cloud_provider::Kind;
+use std::str::FromStr;
+use test_utilities::aws::{K3S_KUBERNETES_MAJOR_VERSION, K3S_KUBERNETES_MINOR_VERSION};
+use test_utilities::common::{cluster_test, ClusterDomain, ClusterTestType};
+
+pub const AWS_K3S_VERSION: &str = "v1.20.15+k3s1";
+
+#[cfg(feature = "test-aws-infra")]
```
benjaminch, line 17: Shouldn't we isolate functional tests for EC2 into their own tests folders with their own features / tests flags?
I guess we might support different K8s versions for AWS and EC2 at some point, having a dedicated folder would avoid mix and dirty plumbering work down the road.

src/cloud_provider/aws/kubernetes/mod.rs
evoxmusic: I just refactored the existing code base. It was already present

src/cloud_provider/aws/kubernetes/mod.rs
```diff
@@ -119,1675 +109,1168 @@ pub struct Options {
...
+                        }
+                    }
+                }
             }
             Err(e) => {
-                // enable cluster autoscaler deployment
-                let _ = self.set_cluster_autoscaler_replicas(event_details.clone(), 1)?;
-
-                return Err(EngineError::new_terraform_error_while_executing_pipeline(event_details, e));
+                let message_safe = "Unable to get helm list";
```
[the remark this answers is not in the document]
evoxmusic: Same I didn't add anything. It was already present

tests/aws/aws_kubernetes_ec2.rs
evoxmusic: It makes sense to me but I don't want to spend too much time on plumbing details until we have something working. If it does not require too much effort - you can structure the way you prefer. Otherwise we can delay it.

evoxmusic:
@benjaminch I didn't comment all your messages since I refactored the existing code and I didn't really add anything new. Feel free to apply your changes directly. You don't need to ask me 😁

src/cloud_provider/aws/kubernetes/mod.rs
benjaminch: ack

src/cloud_provider/aws/kubernetes/mod.rs
benjaminch: ack

src/cloud_provider/aws/kubernetes/mod.rs
benjaminch: FAR

src/cloud_provider/aws/kubernetes/mod.rs
benjaminch: FAR

src/cloud_provider/aws/kubernetes/mod.rs
benjaminch: Done

test_utilities/src/common.rs
benjaminch: Done

tests/aws/aws_kubernetes_ec2.rs
benjaminch: Done

src/cloud_provider/aws/kubernetes/mod.rs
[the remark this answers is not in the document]
benjaminch: Done, fixed into AWS here => #700

src/cloud_provider/aws/kubernetes/mod.rs
benjaminch: Done, fixed into AWS here => #700

PR #696, opened 2022-04-21
feat: add sidecar container to log app start / stop events

benjaminch:
Ticket: ENG-1160


benjaminch:
Canceled, will be handled in a later version eventually.

PR #708, opened 2022-05-05
feat: add qovery dns support

deimosfr:


lib/aws-ec2/bootstrap/tf-default-vars.j2.tf
```diff
@@ -70,7 +70,9 @@ variable "ec2_image_info" {
 variable "ec2_instance" {
   description = "EC2 instance configuration"
   default = {
-    "instance_type" = "t3.micro"
+    "instance_type" = "{{ eks_worker_nodes[0].instance_type }}"
```
benjaminch, line 73: Shouldn't we  avoid reusing `eks_xx` but rather introduce `ec2_xx` ?

lib/aws-ec2/bootstrap/tf-default-vars.j2.tf
```diff
@@ -70,7 +70,9 @@ variable "ec2_image_info" {
...
+    "disk_size_in_gb" = "{{ eks_worker_nodes[0].disk_size_in_gib }}"
```
benjaminch, line 74: Same

erebe (approved):

lib/aws-ec2/bootstrap/chart_values/external-dns.j2.yaml
```diff
@@ -0,0 +1,22 @@
...
+cloudflare:
+  apiToken: "{{ cloudflare_api_token }}"
+  email: "{{ cloudflare_email }}"
+  proxied: {{ cloudflare_proxied|default(value="false") }}
+{% elif external_dns_provider == "pdns" %}
+# Qovery DNS
+pdns:
+  apiUrl: "{{ qoverydns_api_url }}"
+  apiPort: "{{ qoverydns_api_port }}"
+  apiKey: "{{ qoverydns_api_key }}"
```
mzottola, line 13: Seems to be an incoherence with those `qoverydns_` values ?
Here in github there is `qoverydns_` whereas in [gitlab](https://gitlab.com/qovery/engine/qovery-engine/-/merge_requests/255/diffs#77de5415e2e22e2591cd7f70c7895e5c5ff25eac_413_432) part there is `qovery_dns_`

src/cloud_provider/aws/kubernetes/eks.rs
```diff
@@ -669,3 +707,48 @@ impl Listen for EKS {
...
+#[cfg(test)]
+mod tests {
+    use crate::cloud_provider::aws::kubernetes::eks::EKS;
+    use crate::cloud_provider::models::NodeGroups;
+    use crate::errors::Tag;
+    use crate::events::{EventDetails, InfrastructureStep, Stage, Transmitter};
+    use crate::io_models::QoveryIdentifier;
+
+    #[test]
+    fn test_allowed_eks_nodes() {
```
benjaminch, line 720: OPT: Maybe add more test cases:
- test all allowed instances types
- test all not allowed instances types

It's ok to test only one unsupported.

Also maybe if you test all, maybe split into 3 tests:
-  `test_allowed_eks_nodes`
- `test_unallowed_eks_nodes`
- `test_unsupported_eks_nodes`

src/dns_provider/mod.rs
```diff
@@ -1,12 +1,38 @@
...
 
 use crate::io_models::{Context, Domain};
 
 pub mod cloudflare;
 pub mod errors;
+pub mod qoverydns;
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum Kind {
```
benjaminch, line 17: OPT: split `io` from `domain` models avoiding future issues.
CF https://github.com/Qovery/engine/blob/dev/src/cloud_provider/io.rs#L6 and 
https://github.com/Qovery/engine/blob/dev/src/cloud_provider/mod.rs#L51

Domain struct shouldn't derive Serialize and Deserialize.

tests/aws/aws_kubernetes_ec2.rs
```diff
@@ -24,7 +24,7 @@ fn create_and_destroy_aws_ec2_k3s_cluster(
     engine_run_test(|| {
         let region = AwsRegion::from_str(region.as_str()).expect("Wasn't able to convert the desired region");
         let zones = region.get_zones();
-        let cluster_id = generate_cluster_id(region.to_string().as_str());
+        let cluster_id = "ztest-abe3e22b0".to_string(); // don't change it to test qovery dns provider properly
```
benjaminch, line 27: Could it be set into VAULT so it's easier to change it since it's config?

lib/aws-ec2/bootstrap/tf-default-vars.j2.tf
deimosfr: Same, ATM to go fast, it's simpler

lib/aws-ec2/bootstrap/chart_values/external-dns.j2.yaml
deimosfr: updating gitlab side

PR #739, opened 2023-10-03
Update repo URL for Heroku's CNB builder images

edmorley:
Since the repo has just been renamed:
https://github.com/heroku/cnb-builder-images/issues/396


edmorley:
cc @evoxmusic

